PR #12, opened 2022-01-31
Minor improvements to `combine_cookie_crumbs/1`

moogle19:
Replaced the `map` & `join` with a `map_join` and prepend the resulting `combinated_cookie` instead of list concatenation.

### DIFF
--- a/lib/bandit/http2/stream_task.ex
+++ b/lib/bandit/http2/stream_task.ex
@@ -91,9 +91,9 @@ defmodule Bandit.HTTP2.StreamTask do
   defp combine_cookie_crumbs(headers) do
     {crumbs, other_headers} = headers |> Enum.split_with(fn {header, _} -> header == "cookie" end)
 
-    combined_cookie = crumbs |> Enum.map(fn {"cookie", crumb} -> crumb end) |> Enum.join("; ")
+    combined_cookie = Enum.map_join(crumbs, "; ", fn {"cookie", crumb} -> crumb end)
 
-    other_headers ++ [{"cookie", combined_cookie}]
+    [{"cookie", combined_cookie} | other_headers]
   end
 
   defp method(headers), do: get_header(headers, ":method")


lib/bandit/http2/stream_task.ex
```diff
@@ -91,9 +91,9 @@ defmodule Bandit.HTTP2.StreamTask do
   defp combine_cookie_crumbs(headers) do
     {crumbs, other_headers} = headers |> Enum.split_with(fn {header, _} -> header == "cookie" end)
 
-    combined_cookie = crumbs |> Enum.map(fn {"cookie", crumb} -> crumb end) |> Enum.join("; ")
+    combined_cookie = Enum.map_join(crumbs, "; ", fn {"cookie", crumb} -> crumb end)
 
-    other_headers ++ [{"cookie", combined_cookie}]
+    [{"cookie", combined_cookie} | other_headers]
```
mtrudel, line 96: I can't quite recall what the reason was, but I'm pretty sure the appending here is for a specific clause in the RFC. In any case, I certainly wouldn't have written it this way unless there was a definite reason (my personal style is exactly what your suggestion here is). OK to walk this line back?

mtrudel (changes requested):

mtrudel:
Huh. Tests (including h2spec) pass even with the cookie reordering. Lemme take a quick look over the RFC to verify but I think we'll be OK with this PR as-is. Stay tuned.

lib/bandit/http2/stream_task.ex
moogle19: Yeah, no problem. 

Do you remember if it was a http2 specific thing with the header ordering? 
I only found [this](https://datatracker.ietf.org/doc/html/rfc7540#section-8.1.2.5) mentioning cookies in the RFC.

mtrudel:
h2spec tests are green & the relevant RFC (RFC6265§4.2.2) is clear that crumb ordering is not significant. This PR is good to go.

Thanks for the PR!

PR #13, opened 2022-01-31
Avoid regex matching in `should_keepalive?/2`

moogle19:
I replaced the `Regex` matching with a pattern match for a `nil` connection header and otherwise `downcase` the `connection_header` and match it in a case.
This should (in theory) be faster.

### DIFF
--- a/lib/bandit/http1/adapter.ex
+++ b/lib/bandit/http1/adapter.ex
@@ -102,7 +102,7 @@ defmodule Bandit.HTTP1.Adapter do
     end
   end
 
-  # If we do not have a connection header, then keep alive if we're running on HTTP/1.1
+  # If we do not have a connection header, then keep alive iff we're running on HTTP/1.1
   defp should_keepalive?(version, nil), do: version == :"HTTP/1.1"
 
   defp should_keepalive?(version, connection_header) do
--- a/lib/bandit/http1/adapter.ex
+++ b/lib/bandit/http1/adapter.ex
@@ -102,8 +102,8 @@ defmodule Bandit.HTTP1.Adapter do
     end
   end
 
-  defp should_keepalive?(:"HTTP/1.1", nil), do: true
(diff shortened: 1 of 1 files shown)


lib/bandit/http1/adapter.ex
```diff
@@ -102,13 +102,14 @@ defmodule Bandit.HTTP1.Adapter do
     end
   end
 
+  defp should_keepalive?(:"HTTP/1.1", nil), do: true
+  defp should_keepalive?(_, nil), do: false
+
```
mtrudel, line 107:
```suggestion
    # If we do not have a connection header, then keep alive iff we're running on HTTP/1.1
    defp should_keepalive?(version, nil), do: version == :"HTTP/1.1"

```

mtrudel (changes requested):
LGTM - a small readability change only

lib/bandit/http1/adapter.ex
mtrudel: Indent issue - CI will catch that.

lib/bandit/http1/adapter.ex
```diff
@@ -102,8 +102,8 @@ defmodule Bandit.HTTP1.Adapter do
     end
   end
 
-    # If we do not have a connection header, then keep alive iff we're running on HTTP/1.1
-    defp should_keepalive?(version, nil), do: version == :"HTTP/1.1"
+  # If we do not have a connection header, then keep alive if we're running on HTTP/1.1
```
mtrudel, line 105: That wasn't actually a typo - iff == if and only if. No matter, the intent of the comment still stands.

moogle19: Ah okay. Didn't knew that. 
Just reverted it back to "iff".

mtrudel:
Thanks for the PR!

Issue #9, opened 2021-10-12
Advertise HTTP/1.1 via ALPN


mtrudel:
We already advertise h2 (as required by RFC7540), but should also do HTTP/1.1. Configuration of this isn't required for 0.4.x; we'll be adding in broader protocol configuration in 0.6.x

PR #14, opened 2022-01-31
Explicitly handle HTTP1 ALPN

moogle19:
Explicitly handle the `http/1.1` and `http/1.0` ALPN from the client.

### DIFF
--- a/lib/bandit/initial_handler.ex
+++ b/lib/bandit/initial_handler.ex
@@ -23,6 +23,9 @@ defmodule Bandit.InitialHandler do
       {Bandit.HTTP2.Handler, Bandit.HTTP2.Handler} ->
         {:ok, Bandit.HTTP2.Handler, <<>>}
 
+      {Bandit.HTTP1.Handler, {:no_match, data}} ->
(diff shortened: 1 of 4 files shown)


mtrudel:
Great work! I'm very much in favour of this being merged but there are a few loose ends first:

* We already advertise http/1.1, so #9 should be marked as fixed by this PR (I'll take this one)
* We'll need test cases in https://github.com/mtrudel/bandit/blob/main/test/bandit/initial_handler_test.exs to cover newly addressed permutations (we already cover happy paths for both http/1 and h2 in there, but we'll want to add a 'fails to connect via h2 after negotiating http/1.1' case (the opposite of the case on line 34, essentially).

mtrudel:
Fixes #9

moogle19:
@mtrudel 
I added a test case for h2 over HTTP/1.1 and moved the `tls_client` function from the `SimpleH2Client` to `ClientHelpers`  (with `protocols` as a parameter) to reuse it for the test.

mtrudel:
Amazing work! Thanks so much!

PR #16, opened 2022-02-01
Fix `send_file` for chunked files

moogle19:
The `Enum.reduce` in `send_file` returned `:ok` instead of the adapter, so `send_file` would fail for files > 2048 bytes.
I changed it to correctly return the `adapter` and added a test for chunked file sending.

What was the intention on using `Enum.reduce` here?
The return value of `chunk/2` isn't checked and the adapter doesn't change.
Should it be checked? If not wouldn't `Enum.each/2` be a better alternative?


test/bandit/http2/plug_test.exs
```diff
@@ -241,6 +241,20 @@ defmodule HTTP2PlugTest do
...
+
+    assert response.status == 200
+    assert response.body == File.read!(Path.join([__DIR__, "../../support/sendfile_chunked"]))
+  end
+
+  def chunked_file_test(conn) do
+    conn
+    |> send_file(200, Path.join([__DIR__, "../../support/sendfile_chunked"]), 0, :all)
+  end
+
```
mtrudel, line 257: I think the actual intent of this test is to ensure that it does the right thing with 'large' files. Suggest changing the language here to be something like 'sending a file which is larger than 2048 bytes'. I mention this mostly just because 'chunk' has a specific meaning within HTTP and the fact that we use chunking internally is really just an implementation detail that's not really the business of the test. Ultimately we just want to ensure that we get the file read back to us.

lib/bandit/http2/adapter.ex
```diff
@@ -113,7 +114,6 @@ defmodule Bandit.HTTP2.Adapter do
     # (RFC7230§4.1). The whole notion of chunked encoding is moot in HTTP/2 anyway (RFC7540§8.1)
     # so this entire section of the API is a bit slanty regardless.
     send_data(adapter, chunk, chunk == <<>>)
-    :ok
```
mtrudel, line 116: This is as intended. As noted elsewhere on this review, we're constrained in what we can return here by our conformance to the `Plug.Conn.Adapter` behaviour. Specifically, Plug's [explicit guidance](https://hexdocs.pm/plug/Plug.Conn.Adapter.html#c:chunk/2) here is that we return `:ok`, as we currently do. 

I agree that this is an awkward corner of the Plug API, and is one of those rare cases where things aren't entirely functional (nothing actually changes in a `Plug.Conn` struct as a result of a chunk call, even though bytes are sent out on the wire). But we're doing the right thing here already.

lib/bandit/http2/adapter.ex
```diff
@@ -81,6 +81,7 @@ defmodule Bandit.HTTP2.Adapter do
           File.stream!(path, [], 2048)
           |> Enum.reduce(adapter, fn chunk, adapter ->
             chunk(adapter, chunk)
+            adapter
```
mtrudel, line 84: See comments on the larger review - this should be updated to an Enum.each iteration and comments added to the effect that the chunk return value is awkward to use here.

mtrudel (changes requested):
A couple of comments wrt working and existing intent, plus the following.

You're correct - the way we ought to be processing this is to use an `each` in place of `reduce`, and explicitly note in comments that we're using the same `adapter` instance without updating it (since the return value from chunk/2 is so anemic). It should be clear to the reader that this is the case, and using each + comments should make that so.

mtrudel:
Thanks for the PR! Comments in review form are above.

test/bandit/http2/plug_test.exs
moogle19: I updated the naming to make the intention of the test case clearer.

lib/bandit/http2/adapter.ex
moogle19: Oh, sorry. I should have checked the `Plug` documentation first. 
I reverted the change to always return `:ok` here.

lib/bandit/http2/adapter.ex
moogle19: I changed it to `Enum.each/2` and added a comment. 
Please take a quick look if the comment is correct and sufficient enough.

moogle19: I changed it to `Enum.each/2` and added a comment. 
Please take a quick look if the comment is correct and sufficient enough.

mtrudel (approved):
Looks great!

PR #22, opened 2022-09-15
Add Autobahn TestSuite and make it pass

moogle19:
Added a test case for the Autobahn WebSocket-TestSuite and fixed some problems which occurred during this tests:

- Make header check during Upgrade case-insensitive
- Check reserved bits and fail if not 0 (RFC: `MUST be 0 unless an extension is negotiated that defines meanings for non-zero values.`)
- Check close code for validity (not sure why Autobahn expects 1016-2999 to fail but allows 3001-3999 since both are `Unassigned`, see: [IANA](https://www.iana.org/assignments/websocket/websocket.xhtml))
- Check if TextFrame and CloseFrame contain valid UTF-8 string
- Fix masking algorithm
- Append ContinuationFrames to initial Frame (not quite happy with using the initial frame as a buffer for the message, but couldn't think of a better way yet)

With these changes the Autobahn-TestCases 1 - 10 are passed, which would mean the last thing which is missing, is compression.

To run the TestSuite you need Docker. I added the alias `mix test.autobahn` for convenience and don't run the Autobahn Suite by default. 

Please feel free to review the code and propose changes.

### DIFF
--- a/test/bandit/websocket/autobahn_test.exs
+++ b/test/bandit/websocket/autobahn_test.exs
@@ -38,7 +38,7 @@ defmodule WebsocketAutobahnTest do
       end
 
       @impl Sock
-      def handle_ping_frame(ping, socket, state) do
+      def handle_ping_frame(_ping, _socket, state) do
         {:continue, state}
       end
 
@@ -53,14 +53,12 @@ defmodule WebsocketAutobahnTest do
       end
 
       @impl Sock
-      def handle_error(error, _socket, _state) do
-        IO.inspect(error)
+      def handle_error(_error, _socket, _state) do
         :ok
       end
 
       @impl Sock
       def handle_timeout(_socket, _state) do
-        IO.inspect("timeout")
         :ok
       end
 
--- a/test/bandit/websocket/autobahn_test.exs
+++ b/test/bandit/websocket/autobahn_test.exs
@@ -1,0 +1,111 @@
+defmodule WebsocketAutobahnTest do
+  use ExUnit.Case, async: false
+
+  @tag timeout: :infinity
+  @tag :autobahn
+  test "autobahn test suite" do
+    defmodule EchoSock do
+      @behaviour Sock
+
+      import Sock.Socket
+      import Plug.Conn
+
+      @impl Sock
+      def init(_args) do
+        []
+      end
+
+      @impl Sock
+      def negotiate(conn, state) do
+        {:accept, conn, state}
+      end
+
+      @impl Sock
+      def handle_connection(_socket, state) do
+        {:continue, state}
+      end
+
+      @impl Sock
+      def handle_text_frame(text, socket, state) do
+        send_text_frame(socket, text)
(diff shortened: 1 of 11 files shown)


mtrudel:
This looks absolutely superb. All of the items here are things I've had on my todo list (which, now that the project is starting to get some traction I should really turn into public issues for visibility). 

At a first glance, the implementation of everything here looks solid. A couple of items may have a larger follow-on than what's here (like, the changes make sense but there are other places in the code that a similar change should also be made for consistency that isn't reflected here). 

My first thought, in order to get all the bits here over the line in a timely manner, would be to split this PR up into a few smaller ones:

1. One PR to add the autobahn suite & get it running as part of CI (without worrying about failing tests).
2. One PR to add continuation support (there are a few test follow-ons here as well, largely around UTF-8 enforcement as in §5.6
3. Misc small fixes (masking, reserved bits, case insensitivity on headers)
4. mmmmmmmmaybe some deeper thought about error codes. I thought I'd had my head around this one in the RFC pretty well, so it's surprising to see such substantial changes (not that I doubt them at all, I just want to understand them in isolation)

I would propose narrowing the scope of this particular PR to the first point above (ie: get autobahn landed & in CI, don't worry about the failing tests just yet).

WDYT? 

This is *really* great work. Thanks again for the seriously hard work here!

test/bandit/websocket/autobahn_test.exs
```diff
@@ -0,0 +1,109 @@
+defmodule WebsocketAutobahnTest do
+  use ExUnit.Case, async: false
+
+  @tag timeout: :infinity
+  @tag :autobahn
```
mtrudel, line 5: This can likely take a lot of the config conventions from https://github.com/mtrudel/bandit/blob/main/test/bandit/http2/h2spec_test.exs as it's the same conceptual idea (albeit docker only). Specifically:

* Use the external_conformance tag for test inclusion (no test_helper or mix changes needed)
* Use capture_log to keep it quiet in the happy case. If the test fails, ex_unit will dump it all

mtrudel (changes requested):
Small comment on how to isolate autobahn tests. Also, suggestion on the PR proper about scoping this one down to just be autobahn work

moogle19:
@mtrudel 
I just submitted a new PR for the Autobahn CI stuff. 
I will start splitting the rest of this PR up into smaller parts over the weekend.  🙂

mtrudel:
All of the work on this branch has found its way into the 0.5 branch, either by way of #23, #24, or commits directly on the branch (in particular, the continuation, error code, and masking fixes all needed rework beyond what would have feasible to rework on the PR, so I landed those manually). 

I'm about to merge the 0.5 work into main as #26 & will be cutting 0.5.1 off of that. I'll also be creating a bunch of issues here to outline the things still remaining to see websocket support through to a final 0.5.x release. 

Thanks again for the enormous effort here. There was a LOT to get through and you made it look easy.

PR #23, opened 2022-09-16
Add Autobahn CI task

moogle19:
This is the Autobahn part of the #22 PR.

I made the jobs matrix more static, since some of the combinations (e.g. 1.10.4 + OTP 24.2.2) didn't work / exist.
I just took the last 4 versions of Elixir and OTP, I hope that's enough.

I also added a lint flag, so format checking, credo and dialyzer only run once.

Dialyzer also was complaining about `Sock.opts()` type which doesn't exist (yet). I switched it to `any()` for now.


mtrudel:
Merging w/ failing autobahn tests, as they'll be taken care of shortly

PR #24, opened 2022-09-16
Make header field handling case-insensitive

moogle19:
Since header fields are case-insensitive, I added a helper which downcases the fields before comparison.

This is needed, so the Autobahn tests work, since Autobahn uses:
```
Upgrade: WebSocket
Connection: Upgrade
```
in headers


mtrudel:
Looks great (this is mandated in RFC6455§4.2.1.3-4)! 

Added a quick test coverage & merging!

PR #30, opened 2022-09-26
Add date header field

moogle19:
[RFC7231](https://www.rfc-editor.org/rfc/rfc7231#section-7.1.1.2) states the following: 

> An origin server MUST NOT send a Date header field if it does not
   have a clock capable of providing a reasonable approximation of the
   current instance in Coordinated Universal Time.  An origin server MAY
   send a Date header field if the response is in the 1xx
   (Informational) or 5xx (Server Error) class of status codes.  An
   origin server MUST send a Date header field in all other cases.

Since most (if not all) machines which are running bandit are able to get a `reasonable approximation of the
current instance in Coordinated Universal Time` the `Date` header should be added to every Bandit response (if I understand the last line in the RFC correctly).


test/support/simple_websocket_client.ex
```diff
@@ -16,19 +16,23 @@ defmodule SimpleWebSocketClient do
...
     cache-control: max-age=0, private, must-revalidate\r
     upgrade: websocket\r
     connection: Upgrade\r
-    sec-websocket-accept: s3pPLMBiTxaQ9kYGzzhZRbK+xOo=\r
+    sec-websocket-accept: s3pPLMBiTxaQ9kYGzzhZRbK\+xOo=\r
     \r
     """
 
-    {:ok, response} = :gen_tcp.recv(client, byte_size(expected))
+    date_length = byte_size("date: #{Bandit.HTTP.date_header()}\r")
```
moogle19, line 30: This is a bit ugly, but since the day part of the date could be 1 or 2 digits the size of the expected response is not static.

ryanwinchester: Shouldn't it always be the same length?

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Date

ryanwinchester: > day part of the date could be 1 or 2 digits

I believe it should always be 2 digits, which means that it would always be the same length.

lib/bandit/http.ex
```diff
@@ -0,0 +1,9 @@
+defmodule Bandit.HTTP do
+  @moduledoc false
+  # Implements functions shared by different HTTP versions
+
+  # Current DateTime formatted for HTTP headers
+  def date_header do
+    Calendar.strftime(DateTime.utc_now(), "%a, %-d %b %Y %X GMT")
```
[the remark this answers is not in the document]
mtrudel: It doesn't need to be (and depending on your read of the following reference, *shouldn't be*) zero padded (see https://www.rfc-editor.org/rfc/rfc5322#section-3.6.1). The Mozilla docs contradict RFC5322 here; I think the RFC wins.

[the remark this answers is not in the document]
moogle19: Looking at [RFC9110](https://www.rfc-editor.org/rfc/rfc9110#name-date-time-formats) is must be 2 digits.
This RFC seems to be the current one.

mtrudel:
I've been waffling back and forth on adding date headers for a while (not the least because a performant implementation [isn't trivial](https://github.com/ninenines/cowboy/blob/master/src/cowboy_clock.erl)). Truthfully, it's also because this sort of ambiguity in the HTTP/1.x specs drives me nuts (it's a MUST field in some cases, but clients still can't rely on it all of the time because of the unsynced clock carve-out). Thanks for taking it on!

Here's how I think we should structure this to ensure it's performant / well isolated.

* Implement a `Bandit.Clock` module that populates an ets cache with an RFC9110 compliant date string every second
* Provide function(s) that wrap looking this value up so that other callers can consume a clear API (and don't need to sully themselves with ets calls)
* DRY up the test changes (maybe use https://github.com/mtrudel/machete for this?). I'm not nuts about an explosion of regexes in the test suite
* I also think we should make this configurable (we can wait for the 0.6.x branch for this though)

WDYT?

test/support/simple_websocket_client.ex
mtrudel: It's wild that it took like 5 RFCs to get to this. HTTP/1.1 is so underspecified it's scary sometimes.

petermm:
performance wise it might be useful to look at what cowboy does:

[Because querying for the current date and time can be expensive, Cowboy generates one 'Date' header value every second, shares it to all other processes, which then simply copy it in the response. This allows compliance with HTTP/1.1 with no actual performance loss.](https://github.com/ninenines/cowboy/blob/8795233c57f1f472781a22ffbf186ce38cc5b049/doc/src/guide/flow_diagram.asciidoc#date-header)

and handled in the cowboy_clock genserver https://github.com/ninenines/cowboy/blob/master/src/cowboy_clock.erl

mtrudel:
I agree that the broad approach cowboy_clock takes here  should serve as a broad influence.

I don't think we need to be clever about changed parts - generating a single string once per second is a cost we can bear I think :)

By configurable, I just mean that we should allow the insertion of a date field to be optional. For performance reasons, I wonder if we should be doing this check within the clock process (and inserting a sentinel values of some type in the ets table). Better yet, maybe we could put the pre-formed header tuple into ets and just blindly insert it when building header responses (and just populate ets with an empty tuple if we don't want to send the date field). Mostly, I'm thinking about how to do the minimal amount of work necessary on a per-request basis.

lib/bandit/clock.ex
```diff
@@ -0,0 +1,62 @@
...
+  def start_link(_opts) do
+    GenServer.start_link(__MODULE__, [], name: __MODULE__)
+  end
+
+  @impl GenServer
+  def init(_opts) do
+    __MODULE__ = :ets.new(__MODULE__, [:set, :protected, :named_table, {:read_concurrency, true}])
+    update_header()
+    timer_ref = Process.send_after(self(), :update, 1_000)
+    {:ok, %{timer_ref: timer_ref}}
```
mtrudel, line 15: I think this would be better accomplished as a Task with a permanent restart strategy. No need for shutdown handling, no mailbox churn, just a simple tail recursion call.

lib/bandit/http1/adapter.ex
```diff
@@ -289,7 +289,7 @@ defmodule Bandit.HTTP1.Adapter do
   defp response_header(nil, status, headers), do: response_header("HTTP/1.0", status, headers)
 
   defp response_header(version, status, headers) do
-    date_header = "date: #{Bandit.HTTP.date_header()}\r\n"
+    headers = [Bandit.Clock.date_header() | headers]
```
mtrudel, line 292: I wonder if we should be using `Keyword.put_lazy` here in case (for whatever reason) the called Plug has set its own Date header

lib/bandit/http2/adapter.ex
```diff
@@ -150,7 +150,7 @@ defmodule Bandit.HTTP2.Adapter do
 
     headers = [
       {":status", to_string(status)},
-      {"date", Bandit.HTTP.date_header()} | headers
```
mtrudel: ibid.

lib/bandit/clock.ex
```diff
@@ -0,0 +1,62 @@
+defmodule Bandit.Clock do
+  @moduledoc false
+
+  use GenServer
+
+  def start_link(_opts) do
+    GenServer.start_link(__MODULE__, [], name: __MODULE__)
```
mtrudel, line 7: Naming this will preclude running multiple Bandit instances in the same system (which is a specific design goal of Bandit). I think the easiest way to solve this here is to host the Bandit.Clock process within the Bandit application (via `lib/bandit/application.ex` and an update to Bandit's mix.exs). It means that we'd lose the ability to specify date sending on a per-application basis, but I guess that's fine.

mtrudel (changes requested):
Recognizing this is a draft PR, comments are intended in good faith

lib/bandit/http1/adapter.ex
moogle19: `Keyword` wouldn't work here, since the _keys_ are strings.
We could use `if` + `List.keymember?/2`?

mtrudel: Sure!

moogle19: @mtrudel 
The check was a bit more complicated than I thought, since the user defined headers can be any case (e.g. `DaTe`) so I introduced a `has_date_header?/1` function, which uses `Enum.any?/2` with `String.downcase/1`.

lib/bandit/http.ex
```diff
@@ -0,0 +1,13 @@
...
+  @moduledoc false
+  # Implements functions shared by different HTTP versions
+
+  @doc """
+  Checks if a header tuple list already contains a date header
+  """
+  def has_date_header?(headers) do
+    Enum.any?(headers, fn {header, _value} ->
+      String.downcase(header) == "date"
+    end)
```
mtrudel, line 11: This may end up being quite expensive. Can we use an approach like https://hexdocs.pm/plug/Plug.Conn.html#get_resp_header/2 instead (note that we can rely on the `header` being lowercase).

test/bandit/http1/request_test.exs
```diff
@@ -252,6 +257,23 @@ defmodule HTTP1RequestTest do
...
+
+      assert response.status == 200
+
+      assert List.keyfind(response.headers, "date", 0) |> elem(1) ==
+               "Tue, 27 Sep 2022 07:17:32 GMT"
+    end
+
+    def date_header(conn) do
+      conn
+      |> put_resp_header("DaTe", "Tue, 27 Sep 2022 07:17:32 GMT")
```
mtrudel, line 273: Based on comments on https://hexdocs.pm/plug/Plug.Conn.html#put_resp_header/3 I think we can rely on `date` being lowercase, especially if it gains some performance in a hot path.

mtrudel (changes requested):
Looks great overall! A few small changes but almost there!

lib/bandit/clock.ex
```diff
@@ -0,0 +1,44 @@
...
+  the timestamp is newly created for every request
+  """
+  @spec date_header() :: {header :: String.t(), date :: String.t()}
+  def date_header do
+    date =
+      try do
+        :ets.lookup_element(__MODULE__, :date_header, 2)
+      rescue
+        ArgumentError ->
+          get_date_header()
```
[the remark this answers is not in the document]
moogle19: I copied the behaviour from `cowboy_clock`.

To make it more visible I added a warning if the cache isn't available.

lib/bandit/http.ex
moogle19: I updated the `has_date_header?/1` to only check for lowercase date headers.

test/bandit/clock_test.exs
```diff
@@ -0,0 +1,17 @@
+defmodule ClockTest do
+  use ExUnit.Case, async: false
+  import ExUnit.CaptureLog
+
+  test "clock emits warning if ETS cache isn't available" do
+    Application.stop(:bandit)
```
moogle19, line 6: @mtrudel 
Is there a better way to check the warnings than stopping and starting the application?

mtrudel: No, I think this is fine!

lib/bandit/clock.ex
```diff
@@ -0,0 +1,48 @@
...
+  @moduledoc false
+  # Task which updates an ETS table with the current
+  # HTTP header timestamp once a second
+
+  use Task, restart: :permanent
+
+  require Logger
+
+  @doc """
+  Returns the current timestamp according to RFC9110 5.6.7.
```
mtrudel, line 11:
```suggestion
  Returns the current timestamp according to RFC9110§5.6.7.
```
```suggestion
  Returns the current timestamp according to RFC9110 5.6.7.
```

lib/bandit/clock.ex
```diff
@@ -0,0 +1,48 @@
...
+  the timestamp is newly created for every request
+  """
+  @spec date_header() :: {header :: String.t(), date :: String.t()}
+  def date_header do
+    date =
+      try do
+        :ets.lookup_element(__MODULE__, :date_header, 2)
+      rescue
+        ArgumentError ->
+          Logger.warning("Header timestamp couldn't get fetched from ETS cache.")
```
mtrudel, line 23:
```suggestion
          Logger.warning("Header timestamp couldn't be fetched from ETS cache")
```

test/bandit/clock_test.exs
```diff
@@ -0,0 +1,17 @@
...
+
+    warnings =
+      capture_log(fn ->
+        Bandit.Clock.date_header()
+      end)
+
+    assert warnings =~ "Header timestamp couldn't get fetched from ETS cache."
```
mtrudel, line 13:
```suggestion
    assert warnings =~ "Header timestamp couldn't be fetched from ETS cache"
```

test/bandit/clock_test.exs
```diff
@@ -0,0 +1,17 @@
+defmodule ClockTest do
+  use ExUnit.Case, async: false
+  import ExUnit.CaptureLog
+
+  test "clock emits warning if ETS cache isn't available" do
+    Application.stop(:bandit)
+
+    warnings =
+      capture_log(fn ->
+        Bandit.Clock.date_header()
```
mtrudel, line 10:
```suggestion
        date = Bandit.Clock.date_header()
        assert valid_date_header?(date)
```

mtrudel (approved):

mtrudel:
Looks amazing! I made a few super minor changes to copy, just for consistency with the rest of the app. Once CI comes back green I'll merge & release.

mtrudel:
Accidentally merged w/o squashing. Squashed and force pushed to main to fix

mtrudel:
Thanks again @moogle19  for the amazing work!

PR #31, opened 2022-09-29
Minor performance improvements

moogle19:
- Use `String.downcase/2` in `:ascii` mode
- Boolean `and` instead of `&&` in `get_http_protocol` (since all elements are booleans)
- Simplify `do_read_chunk` by using `:binary.split/2` instead of `:binary.match/2` and binary pattern matching


mtrudel:
Great call on the chunked read; that's some very early Bandit code and this is a great improvement. 

Thanks, once again, for a superb PR!

PR #33, opened 2022-10-03
Remove HTTP/2 Push Promise support since Plug is deprecating it

mtrudel:
See https://github.com/elixir-plug/plug/commit/cd258e7acddf8e2cbed62a13c776d63cc17905d8

@moogle19 you've had a lot of eyes here lately; care to give this a once over?


moogle19:
Can't we get rid of the `PushPromise` frame altogether and just handle it as an `Unknown` frame? 

Alternatively if you want to keep the frame, it can be an empty struct since the fields aren't used anyway (as far as I can see). If we want to ignore invalid `PushPromise` frames (zero stream_id and invalid padding length) the deserialization can also be skipped and `Frame.deserialize/2` can return the empty `PushPromise` struct directly.

WDYT?

mtrudel:
> Can't we get rid of the `PushPromise` frame altogether and just handle it as an `Unknown` frame?

Sadly we cannot, since a number of tests in h2spec require us to handle push frames in a specific way (even if we don't support them). We can drop the deserialization down to a single error clause though, since all of the failure cases in the spec resolve down to the same error (protocol error, close connection).

We will need to keep the deserialize function around, but no longer need to have a struct definition for PushPromise.

> Edit: And the `Stream.stream_id_is_valid_server/1` is reported as unused by the compiler.

✅ thanks!

mtrudel:
Updated

lib/bandit/http2/frame/settings.ex
```diff
@@ -33,10 +33,10 @@ defmodule Bandit.HTTP2.Frame.Settings do
         {:cont, {:ok, %{acc | header_table_size: value}}}
 
       {:ok, {0x02, 0x01}}, {:ok, acc} ->
-        {:cont, {:ok, %{acc | enable_push: true}}}
+        {:cont, {:ok, acc}}
```
moogle19, line 36: Can we remove these cases all together (and ignore invalid `enable_push` values)?
`h2spec` seems fine with it.

Otherwise group the valid cases to `{:ok, {0x02, val}}, {:ok, acc} when val in [0x00, 0x01] ->`?

moogle19 (commented):
Just one little thing.
Otherwise it looks good!

lib/bandit/http2/frame/settings.ex
mtrudel: I'd like to keep the handling, even if h2spec doesn't complain. No sense in walking back strict conformance. 

Updated to use a when clause, as suggested

moogle19: Or maybe use `{:ok, {0x02, value}}, {:ok, _acc} when value not in [0x00, 0x01]` on the error case and drop the noop case?

PR #34, opened 2022-10-03
Implement no-op support for Plug.Conn.Adapter.upgrade/3

mtrudel:
This work does not provide any concrete support for any protocols, it merely satisfies the plumbing requirements introduced in Plug by https://github.com/elixir-plug/plug/pull/1119

See https://github.com/phoenixframework/phoenix/issues/5003 for an omnibus overview.

Doing a no-op implementation here allows for the changes required in Plug to be isolated to a small set of PRs (namely this one and https://github.com/elixir-plug/plug_cowboy/pull/88), while not breaking anything in the process.

Note that this PR is not entirely complete yet; it will need to have its dependency on Plug updated to a version which incorporates https://github.com/elixir-plug/plug/pull/1119.


mtrudel:
Closing in favour of #38

PR #36, opened 2022-10-16
Remove un-necessary Enum.map/2

sabiwara:
Hi,

This is a small suggestion to remove some calls to `Enum.map/2` to avoid intermediate structures when there is a way of doing it in one pass.
While it most probably won't result in a noticeable speedup, it also simplifies the code.

The second unrelated commit just adds syntax highlighting to some code in the readme.

### DIFF
--- a/lib/bandit/http2/stream_collection.ex
+++ b/lib/bandit/http2/stream_collection.ex
@@ -35,13 +35,15 @@ defmodule Bandit.HTTP2.StreamCollection do
     delta = initial_send_window_size - collection.initial_send_window_size
 
     streams =
-      Map.new(collection.streams, fn
+      collection.streams
+      |> Enum.map(fn
         {id, %Stream{state: state} = stream} when state in [:open, :remote_closed] ->
           {id, %{stream | send_window_size: stream.send_window_size + delta}}
 
         {id, stream} ->
           {id, stream}
       end)
+      |> Map.new()
 
     %{collection | streams: streams, initial_send_window_size: initial_send_window_size}
   end
--- a/lib/bandit/http2/stream_collection.ex
+++ b/lib/bandit/http2/stream_collection.ex
@@ -35,15 +35,13 @@ defmodule Bandit.HTTP2.StreamCollection do
     delta = initial_send_window_size - collection.initial_send_window_size
 
     streams =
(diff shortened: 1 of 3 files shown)


lib/bandit/http2/stream_collection.ex
```diff
@@ -35,15 +35,13 @@ defmodule Bandit.HTTP2.StreamCollection do
...
-      collection.streams
-      |> Enum.map(fn
+      Map.new(collection.streams, fn
         {id, %Stream{state: state} = stream} when state in [:open, :remote_closed] ->
           {id, %{stream | send_window_size: stream.send_window_size + delta}}
 
         {id, stream} ->
           {id, stream}
       end)
-      |> Map.new()
```
mtrudel, line 46: This ends up as the exact same implementation internally (see https://github.com/elixir-lang/elixir/blob/a64d42f5d3cb6c32752af9d3312897e8cd5bb7ec/lib/elixir/lib/map.ex#L217).

I have a mild stylistic preference for the existing implementation (they're both fine! it's just a matter of convention in that I find explicit maps to be more legible, and it's also the convention used elsewhere in the project.

lib/bandit/websocket/handshake.ex
```diff
@@ -37,9 +37,10 @@ defmodule Bandit.WebSocket.Handshake do
   end
 
   defp header_contains?(conn, field, value) do
+    value = String.downcase(value, :ascii)
+
     conn
     |> get_req_header(field)
-    |> Enum.map(&String.downcase(&1, :ascii))
-    |> Enum.member?(String.downcase(value, :ascii))
+    |> Enum.any?(&(String.downcase(&1, :ascii) == value))
```
mtrudel, line 44: This is good! There's been a lot of (fantastic!) bike shedding over this one function lately; it's cool to see.

mtrudel (changes requested):
All of this PR except for the Map.new change is good to go! Thanks for the PR!

lib/bandit/http2/stream_collection.ex
sabiwara: Sure! I reverted this one.

mtrudel:
Thanks for the PR!

PR #38, opened 2022-10-16
Implement WebSocket support based on Plug.Conn.Adapter.upgrade/3

mtrudel:
This PR provides support for upgrading HTTP connections to WebSocket connections by use of the `c:Plug.Conn.Adapter.upgrade/3` callback. WebSocket implementations are called based on the Sock API (which mirrors `Phoenix.Socket.Transport`) . This work is the Bandit analog of https://github.com/elixir-plug/plug/pull/1119

See https://github.com/phoenixframework/phoenix/issues/5003 for an omnibus overview of the larger body of work that this PR is part of.


mtrudel:
@moogle19 - would you be able to take a look at this in the context of https://github.com/phoenixframework/phoenix/issues/5003?

lib/bandit/websocket/socket.ex
```diff
@@ -1,56 +1,48 @@
 defprotocol Bandit.WebSocket.Socket do
   @moduledoc false
   #
-  # A protocol adding close behaviour to the Sock.Socket protocol
+  # A protocol adding defining the low-level functionality of a WebSocket
```
moogle19, line 4:
```suggestion
  # A protocol defining the low-level functionality of a WebSocket
```

lib/bandit/websocket/handler.ex
```diff
@@ -1,31 +1,32 @@
...
+          timeout -> {:continue, Map.put(state, :connection, connection), {:persistent, timeout}}
+        end
 
-    Connection.handle_connection(state.conn, socket, connection)
-    |> case do
-      {:continue, connection} -> {:continue, update_state(state, connection)}
-      {:continue, connection, timeout} -> {:continue, update_state(state, connection), timeout}
-      {:close, _} -> {:close, state}
+      {:error, reason, connection} ->
+        {:error, reason, state |> Map.put(:connection, connection)}
```
moogle19, line 26:
```suggestion
        {:error, reason, Map.put(state, :connection, connection)}
```
to be consistent with the previous cases

lib/bandit/websocket/connection.ex
```diff
@@ -53,7 +29,7 @@ defmodule Bandit.WebSocket.Connection do
 
       %Frame.Text{fin: true} = frame ->
         if String.valid?(frame.data) do
-          connection.sock.handle_text_frame(frame.data, socket, connection.sock_state)
+          connection.sock.handle_in({frame.data, opcode: :text}, connection.sock_state)
```
moogle19, line 32: More of a `Sock` issue, but what is the intention of using a keyword list as second tuple argument and not just the opcode?

moogle19 (commented):
Apart from a few cosmetic changes this looks awesome! 🙂

lib/bandit/websocket/connection.ex
mtrudel: It's a concession to parity with the existing `Phoenix.Socket.Transport` behaviour (specifically, https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/socket/transport.ex#L163-L200). I agree that it's a bit awkward (especially as it doesn't have parity with the shape of return values in `{:push, ...` and `{:reply, ...` tuples), but it allows Phoenix to serve `Phoenix.Socket` instances as valid, conforming Sock instances.

mtrudel:
Thanks for the review @moogle19 !

PR #41, opened 2022-11-04
Fix header parsing in Websocket upgrade

moogle19:
The Websocket upgrade was failing, when the connection header had more than 1 value (e.g `Connection: keep-alive, Upgrade` which is used by the Phoenix live_reload socket). 
I updated the `header_contains?/3` to handle these cases.

### DIFF
--- a/lib/bandit/websocket/handshake.ex
+++ b/lib/bandit/websocket/handshake.ex
@@ -116,6 +116,7 @@ defmodule Bandit.WebSocket.Handshake do
 
     conn
     |> get_req_header(field)
+    |> Enum.flat_map(&Plug.Conn.Utils.list/1)
     |> Enum.any?(&(String.downcase(&1, :ascii) == value))
   end
 end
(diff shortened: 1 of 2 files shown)


mtrudel:
Good catch (I've tested against live reload in the past; I wonder if this changed upstream recently? What phoenix are you on?)

Fix looks right on the money. Will merge once CI passes.

moogle19:
I was using the `master` branch. Maybe they changed something reacently.

mtrudel:
Thanks for the PR @moogle19 !

mtrudel:
Published as 0.5.11

mtrudel:
What browser was this?

moogle19:
Firefox.
I just checked Chrome and Safari and they don't send the `keep-alive` in the `Connection` header.

mtrudel:
Yeah, I was going to say, neither Chrome or Safari did so (they were the only two browsers I'd tested on). 

It's a bit of a bizarre combination, though I suppose it's technically valid. Anyway, mystery solved.

PR #42, opened 2022-11-10
Gracefully handle invalid HTTP versions

moogle19:


mtrudel:
Nice!

lib/bandit/http1/adapter.ex
```diff
@@ -75,13 +75,19 @@ defmodule Bandit.HTTP1.Adapter do
...
+              %{req | buffer: rest, version: version},
+              :httph,
+              headers,
+              method,
+              path
+            )
+
+          {:error, reason} ->
+            {:error, reason}
+        end
```
ryanwinchester, line 90:
```suggestion
        with {:ok, version} <- get_version(version) do
          do_read_headers(%{req | buffer: rest, version: version}, :httph, headers, method, path)
        end
```

mtrudel: Nice golfing!

mtrudel: Since this is already merged I'll bring this in on main (can't credit you though :() Unless you want to spin up a PR?

ryanwinchester: I can do it. There is a bunch of the same `case` types we could do in the same module, now that I'm looking at it. So, for consistency I can change them all or we just don't touch it it all?

mtrudel: Do em all!

PR #43, opened 2022-11-10
Handle all path types

moogle19:
Currently only the `:abs_path` returned by `:erlang.decode_packet/3` are handled.
This adds support for the other types by extracting the `path` component of the [HttpUri](https://www.erlang.org/doc/man/erlang.html#decode_packet-3) returned by `decode_packet/3`.

Because of the URL construction [here](https://github.com/mtrudel/bandit/blob/f01d156b9b10cf51b74eaea4938361097d50f9f4/lib/bandit/http1/handler.ex#L42), the path **must** start with a slash, otherwise that will fail, therefore a slash is added to the start of the path if it is not present.

I am not sure if this is the correct or best way to handle the paths, so feedback would be appreciated!

### DIFF
--- a/lib/bandit/http1/adapter.ex
+++ b/lib/bandit/http1/adapter.ex
@@ -129,7 +129,7 @@ defmodule Bandit.HTTP1.Adapter do
   defp resolve_path(_path),
     do:
       {:error,
-       "Not supported. Path must be an absolute path, an absolute URI or `*` for a global OPTIONS request."}
+       "Not supported. Path must be an absolute path, an absolute URI or `*` request."}
 
   ##############
   # Body Reading
--- a/lib/bandit/http1/adapter.ex
+++ b/lib/bandit/http1/adapter.ex
@@ -72,7 +72,7 @@ defmodule Bandit.HTTP1.Adapter do
 
       {:ok, {:http_request, method, path, version}, rest} ->
         with {:ok, version} <- get_version(version),
-             {:ok, path} <- resolve_path(path, method) do
+             {:ok, path} <- resolve_path(path) do
           do_read_headers(%{req | buffer: rest, version: version}, :httph, headers, method, path)
         end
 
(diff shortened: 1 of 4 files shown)


mtrudel:
Are such requests specifically allowed by RFC? I'm old enough to remember the days when Apache had all manner of security issues caused by relative URI parsing and it's not a can of worms that I'm keen to open up unless we have to.

moogle19:
[RFC9112](https://www.rfc-editor.org/rfc/rfc9112#request.target) should be the current one, but the allowed paths are the same.

[RFC9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-determining-the-target-reso) also states:
- For CONNECT ([Section 9.3.6](https://www.rfc-editor.org/rfc/rfc9110.html#CONNECT)), the request target is the host name and port number of the tunnel destination, separated by a colon.
- For OPTIONS ([Section 9.3.7](https://www.rfc-editor.org/rfc/rfc9110.html#OPTIONS)), the request target can be a single asterisk ("*").

See the respective method definitions for details. These forms **MUST NOT** be used with other methods.

mtrudel:
To summarize, I think we should have the following behaviour?:

* if the request line is an absolute path `/path/to/a/thing?a=b`, then
    * path: `/path/to/a/thing`
    * query string: `a=b`
    * host as defined in Host header

* if the request line is an absolute URI `http://example.com:8080/path/to/a/thing?a=b`, then
    * path: `/path/to/a/thing`
    * query string: `a=b`
    * host is example.com (RFC2616§5.2.3 says to ignore Host)
 
* if the request line is `*`, then:
    * path: `*`
    * host as defined in Host header
    * Should we return a 400 if method is not OPTIONS (I don't think so, since the specs are loose on this)
    * host as defined in Host header

* If the request line is authority `example.com:8080`, then
  * path: ????
  * host is example.com
  * Port is required per RFC9110§9.3.6

* I don't think we should support relative URIs at all. I don't think any spec references them and it feels like a Pandora's box.

WDYT?

moogle19:
[The RFC9112 states](https://www.rfc-editor.org/rfc/rfc9112#section-3.2.4): 
> The "asterisk-form" of request-target is only used for a server-wide OPTIONS request ([Section 9.3.7](https://www.rfc-editor.org/rfc/rfc9110#section-9.3.7) of [[HTTP](https://www.rfc-editor.org/rfc/rfc9112#HTTP)]).

Returning 400 would be okay in that case.

For the other cases I'm okay with your proposed behaviour.
I'll update PR accordingly.

ryanwinchester:
> Yeah but it's never actually called out in screaming case MUST, which is notable (there are also places in 2616 where the language is even looser). My vote would be to push this down to the implementing plug and keep the policy out of bandit.

@mtrudel I think you edited their comment instead of replying ^

mtrudel:
So I did! I was on mobile at the time, I guess that's my excuse LOL.

@moogle19 tagging you in case you did t get notified for the 'reply' above

moogle19:
@mtrudel 
Updated the PR.

I decided return an error on `:scheme` / authority paths (I am a bit confused by the erlang naming, since for path like `example.com:8080`, `example.com` would be the scheme and `80` the path).
As it seems to be only used by `CONNECT` requests and we don't support tunneling, I think we can ignore it.

mtrudel:
Pushed up a couple of small cleanup changes:
* renamed 'path' to 'request_target' for cleaner semantics
* tidied up a couple of (unrelated) function calls to keep formatted code sane
* tidied up `:*` use in tests to avoid having to special case it

LMK if the changes look good to you then I think we can merge!

moogle19:
Looks good! 👍

PR #47, opened 2022-11-10
Convert `case` with passthrough errors to `with` in `Bandit.HTTP1.Adapter`

ryanwinchester:
🏌️ As discussed https://github.com/mtrudel/bandit/pull/42#discussion_r1019619895


mtrudel:
Nice! Thanks for the PR!

PR #48, opened 2022-11-11
Fix Elixir 1.15 deprecations / warnings

moogle19:
While test-driving 1.15 I noticed some warnings.

`Logger.warn/2` will be deprecated in favor of `Logger.warning/2`. 
As it is available since `1.11` and Bandit defines `~> 1.11` as elixir version this shouldn't be a problem.

And `mix format` defaults to `~c` for charlist now.


mtrudel:
A M A Z I N G

PR #53, opened 2022-11-17
Handle negative `content-length` header

moogle19:
Currently a negative `content-length` header results in the following error:
```

11:11:13.109 [info] Running MyPlug with Bandit 0.6.0 at 0.0.0.0:4000 (http)

11:11:16.385 [error] GenServer #PID<0.383.0> terminating
** (FunctionClauseError) no function clause matching in :prim_inet.recv0/3
    :prim_inet.recv0(#Port<0.11>, -328, 15000)
    (thousand_island 0.5.11) lib/thousand_island/socket.ex:78: ThousandIsland.Socket.recv/3
    (bandit 0.6.0) lib/bandit/http1/adapter.ex:218: Bandit.HTTP1.Adapter.grow_buffer/4
    (bandit 0.6.0) lib/bandit/http1/adapter.ex:160: Bandit.HTTP1.Adapter.read_req_body/2
    (plug 1.14.0) lib/plug/conn.ex:1158: Plug.Conn.read_body/2
    #cell:wxlzqfftjgta7e6vzfyvswxlroajtju3:7: MyPlug.call/2
    (bandit 0.6.0) lib/bandit/http1/handler.ex:73: Bandit.HTTP1.Handler.call_plug/2
    (bandit 0.6.0) lib/bandit/http1/handler.ex:15: Bandit.HTTP1.Handler.handle_data/3
State: {%ThousandIsland.Socket{socket: #Port<0.11>, transport_module: ThousandIsland.Transports.TCP, connection_id: "E7BD30C0B1D5", acceptor_id: "0C74D2FB7D59", read_timeout: 15000}, %{handler_module: Bandit.InitialHandler, plug: {MyPlug, []}}}
```

Because the negative `body_remaining` is used as a param for `Socket.recv/3`.
~~This will terminate the reading of the body, when the `body_remaining` is less or equal to `0`.~~


lib/bandit/http1/adapter.ex
```diff
@@ -109,6 +112,21 @@ defmodule Bandit.HTTP1.Adapter do
   defp get_version({1, 0}), do: {:ok, :"HTTP/1.0"}
   defp get_version(other), do: {:error, "invalid HTTP version: #{inspect(other)}"}
 
+  defp get_content_length(headers) do
+    with {_, value} <- List.keyfind(headers, "content-length", 0),
+         [length] <- Enum.uniq(Plug.Conn.Utils.list(value)),
```
moogle19, line 120: Related text in the [RFC9112](https://www.rfc-editor.org/rfc/rfc9112.html#section-6.3)
> If a message is received without [Transfer-Encoding](https://www.rfc-editor.org/rfc/rfc9112.html#field.transfer-encoding) and with an invalid [Content-Length](https://www.rfc-editor.org/rfc/rfc9112.html#body.content-length) header field, then the message framing is invalid and the recipient MUST treat it as an unrecoverable error, unless the field value can be successfully parsed as a comma-separated list ([Section 5.6.1](https://www.rfc-editor.org/rfc/rfc9110#section-5.6.1) of [[HTTP](https://www.rfc-editor.org/rfc/rfc9112.html#HTTP)]), all values in the list are valid, and all values in the list are the same (in which case, the message is processed with that single value used as the Content-Length field value)

lib/bandit/http1/adapter.ex
```diff
@@ -58,6 +60,10 @@ defmodule Bandit.HTTP1.Adapter do
                connection: connection,
                keepalive: keepalive
            }}
+
```
moogle19, line 63: Related text in [RFC9112](https://www.rfc-editor.org/rfc/rfc9112)

> If a message is received with both a [Transfer-Encoding](https://www.rfc-editor.org/rfc/rfc9112.html#field.transfer-encoding) and a [Content-Length](https://www.rfc-editor.org/rfc/rfc9112.html#body.content-length) header field, the Transfer-Encoding overrides the Content-Length. Such a message might indicate an attempt to perform request smuggling ([Section 11.2](https://www.rfc-editor.org/rfc/rfc9112.html#request.smuggling)) or response splitting ([Section 11.1](https://www.rfc-editor.org/rfc/rfc9112.html#response.splitting)) and ought to be handled as an error. An intermediary that chooses to forward the message MUST first remove the received Content-Length field and process the Transfer-Encoding (as described below) prior to forwarding the message downstream.

moogle19:
After thoroughly reading RFC9112§6.3 I realized this is quite complex. 
I converted this to a draft, since it needs some more work to handle all case defined [here](https://www.rfc-editor.org/rfc/rfc9112.html#name-message-body-length).

lib/bandit/http1/adapter.ex
```diff
@@ -138,9 +160,10 @@ defmodule Bandit.HTTP1.Adapter do
   def read_req_body(%__MODULE__{state: :no_body} = req, _opts), do: {:ok, <<>>, req}
 
   def read_req_body(
-        %__MODULE__{state: :headers_read, buffer: buffer, body_remaining: 0} = req,
+        %__MODULE__{state: :headers_read, buffer: buffer, body_remaining: body_remaining} = req,
         _opts
-      ) do
+      )
+      when body_remaining <= 0 do
```
mtrudel, line 166: I'm not sure that this change is required?

lib/bandit/http1/adapter.ex
```diff
@@ -16,6 +16,7 @@ defmodule Bandit.HTTP1.Adapter do
             upgrade: nil
 
   alias ThousandIsland.Socket
+  alias Plug.Conn.Utils, as: PlugUtils
```
mtrudel, line 19: For the one use of this let's just use it directly below?

lib/bandit/http1/adapter.ex
```diff
@@ -29,27 +30,29 @@ defmodule Bandit.HTTP1.Adapter do
...
            do_read_headers(req) do
-      body_size = get_header(headers, "content-length")
       body_encoding = get_header(headers, "transfer-encoding")
       connection = get_header(headers, "connection")
       keepalive = should_keepalive?(version, connection)
 
-      case {body_size, body_encoding} do
-        {nil, nil} ->
+      case {get_content_length(headers), body_encoding} do
+        {{:ok, nil}, nil} ->
```
mtrudel, line 38: can we use a clause in the above `with` to match on `{:ok, body_size} <- get_content_length(headers)`? I think that would simplify the case matches here (specifically you could drop the nested tuples and also the `{:error...` clause).

lib/bandit/http1/adapter.ex
mtrudel: nice

mtrudel (changes requested):
A couple of suggested changes. Sorry it took me so long to get to this. Life &c.

I think the only other thing missing here would be to handle the differing response codes as you point out. I *think* you could handle those in the `get_content_length` function (assuming you provide it with the response code). That would keep all the logic contained.

mtrudel:
Heads up that I just merged https://github.com/mtrudel/bandit/pull/57 which caused a bunch of diff noise in tests. You'll probably want to rebase on top of that before you get too far along on tests.

lib/bandit/http1/adapter.ex
```diff
@@ -109,6 +116,21 @@ defmodule Bandit.HTTP1.Adapter do
   defp get_version({1, 0}), do: {:ok, :"HTTP/1.0"}
   defp get_version(other), do: {:error, "invalid HTTP version: #{inspect(other)}"}
 
+  defp get_content_length(headers) do
+    with {_, value} <- List.keyfind(headers, "content-length", 0),
+         [length] <- Enum.uniq(PlugUtils.list(value)),
+         length <- String.to_integer(length) do
```
ryanwinchester, line 122: This could be `=`

```suggestion
         length = String.to_integer(length) do
```

mtrudel: We should be using `Integer.parse/1` here to handle cases without raising. See https://github.com/mtrudel/bandit/pull/56/files#diff-d101e20286894b6a1debc54099265d0118e6edd96abca2dbcd2b7febe7db638fR93-R96

(I wonder if we should be factoring that out into a helper, along with a few of our String.split calls &c?)

moogle19:
> A couple of suggested changes. Sorry it took me so long to get to this. Life &c.
> 
> I think the only other thing missing here would be to handle the differing response codes as you point out. I _think_ you could handle those in the `get_content_length` function (assuming you provide it with the response code). That would keep all the logic contained.

If I understand that correctly, the status code are only relevant in the **response** to the client, not the request to the server.
So we only need to look at 6.3.{3-7}, which should be covered by this PR.

lib/bandit/http1/adapter.ex
moogle19: Switched it to `Integer.parse/2` only accepting value without a remainder.

lib/bandit/http1/adapter.ex
moogle19: Yes, but credo was complaining, so i aliased it.
I added a `disable` for `AliasUsage` to silence the warning.

lib/bandit/http1/adapter.ex
moogle19: Nope, it wasn't anymore.

test/bandit/http1/request_test.exs
```diff
@@ -329,6 +329,41 @@ defmodule HTTP1RequestTest do
...
 
+    test "reads a content-length with multiple content-lengths encoded body properly", context do
+      {:ok, response} =
+        Finch.build(
+          :post,
+          context[:base] <> "/expect_body_with_multiple_content_length",
+          [{"connection", "close"}, {"content-length", "8000000,8000000,8000000"}],
+          String.duplicate("a", 8_000_000)
+        )
+        |> Finch.request(context[:finch_name])
```
mtrudel, line 340: Maybe add a test to cover the case where they're comma separated, but different from one another? The twists and turns of the RFC are pretty nasty and there's no better time to pile on the tests than now when we understand them.

Also maybe drop the relevant RFC section in a comment to remember why we're testing them :)

mtrudel (changes requested):
Looks great! Just a quick note about tests

test/bandit/http1/request_test.exs
moogle19: Added a test and a comment to reference the RFC

lib/bandit/http1/adapter.ex
```diff
@@ -106,6 +112,21 @@ defmodule Bandit.HTTP1.Adapter do
   defp get_version({1, 0}), do: {:ok, :"HTTP/1.0"}
   defp get_version(other), do: {:error, "invalid HTTP version: #{inspect(other)}"}
 
+  defp get_content_length(headers) do
+    with {_, value} <- List.keyfind(headers, "content-length", 0),
+         [length] <- Enum.uniq(Plug.Conn.Utils.list(value)),
```
mtrudel, line 117: (Apologies for the code golf here but this is a *really* hot path and I want to make sure it's as fast as possible).

I wonder if we should be special-casing the (almost universally common) case where we have a single content-length value, and avoiding the uniq call. I looked into the Enum.uniq implementation and it's (necessarily) complex, even for lists of length 1. Maybe something like 
```suggestion
         {:ok, length} <- enforce_unique_value(value),
```
along with:
```elixir
defp enforce_unique_value([value]), do: {:ok, value}
defp enforce_unique_value(value) do
  case Enum.uniq(Plug.Conn.Utils.list(value)) do
    [value] -> {:ok, value}
    _ -> {:error, "???"}
  end
end
```

I ran a quick benchee to see the difference, and it's quite staggering:

```
defmodule Foo do
  def uniq do
    [1] 
    |> Enum.uniq 
    |> case do
      [x] -> x
    end
  end

  def single do
    [1] 
    |> case do
      [x] -> x
    end
  end
end

Mix.install([{:benchee, "~> 1.0", only: :dev}])

Benchee.run(%{
  uniq: &Foo.uniq/0,
  single: &Foo.single/0
})


Name             ips        average  deviation         median         99th %
single      338.16 M        2.96 ns   ±334.86%        2.96 ns        3.08 ns
uniq         11.22 M       89.11 ns ±25896.88%          83 ns          84 ns

Comparison:
single      338.16 M
uniq         11.22 M - 30.13x slower +86.15 ns
```




WDYT?

moogle19: I updated the PR and removed the call to `Enum.uniq/1` completely in favor of just checking the values recursively (since it can only be a list). If it is only one value, the function returns early.
We still need the call to [Plug.Conn.Utils.list/1](https://github.com/elixir-plug/plug/blob/d7940552e365e0d3245bb78ca99429ea5c324f7d/lib/plug/conn/utils.ex#L270) though, which could impact performance.

mtrudel:
Merged main in, fixed a minor typo and a racy test. Once CI is green we can merge!

A M A Z I N G

PR #56, opened 2022-11-19
Manually build URI instead of using URI.parse/1 in HTTP/1 adapter

mtrudel:
For improved speed / memory use

### DIFF
--- a/lib/bandit/http1/handler.ex
+++ b/lib/bandit/http1/handler.ex
@@ -29,21 +29,14 @@ defmodule Bandit.HTTP1.Handler do
   end
 
   defp build_conn(req) do
-    case Adapter.read_headers(req) do
-      {:ok, headers, method, request_target, req} ->
-        %{address: remote_ip} = Adapter.get_peer_data(req)
-
-        # Parse a string to build a URI struct. This is quite a hack. In general, canonicalizing
-        # URIs is a delicate process & rather than building a half-baked implementation here it's
-        # better to leave a simple and ugly hack in place so that future improvements are obvious.
-        # Future paths here are discussed at https://github.com/elixir-plug/plug/issues/948)
-        {"host", host} = List.keyfind(headers, "host", 0, {"host", nil})
-        scheme = if Adapter.secure?(req), do: :https, else: :http
-
-        uri = build_uri(scheme, host, request_target)
-
-        {:ok, Plug.Conn.Adapter.conn({Adapter, req}, method, uri, remote_ip, headers)}
-
+    with {:ok, headers, method, request_target, req} <- Adapter.read_headers(req),
+         {:ok, scheme} <- determine_scheme(request_target, req),
+         {:ok, host, port} <- determine_host_and_port(request_target, headers, req),
+         {:ok, path, query} <- determine_path_and_query(request_target),
+         uri <- %URI{scheme: scheme, host: host, port: port, path: path, query: query},
+         %{address: remote_ip} <- Adapter.get_peer_data(req) do
+      {:ok, Plug.Conn.Adapter.conn({Adapter, req}, method, uri, remote_ip, headers)}
+    else
       {:error, :timeout} ->
         attempt_to_send_fallback(req, 408)
         {:error, "timeout reading request"}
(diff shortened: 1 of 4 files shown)


mtrudel:
@moogle19 wanna take a gander?

lib/bandit/http1/handler.ex
```diff
@@ -29,21 +29,14 @@ defmodule Bandit.HTTP1.Handler do
...
-
-        uri = build_uri(scheme, host, request_target)
-
-        {:ok, Plug.Conn.Adapter.conn({Adapter, req}, method, uri, remote_ip, headers)}
-
+    with {:ok, headers, method, request_target, req} <- Adapter.read_headers(req),
+         {:ok, scheme} <- determine_scheme(request_target, req),
+         {:ok, host, port} <- determine_host_and_port(request_target, headers, req),
+         {:ok, path, query} <- determine_path_and_query(request_target),
+         uri <- %URI{scheme: scheme, host: host, port: port, path: path, query: query},
```
ryanwinchester, line 36: Since I'm bringing up this in the other PR, could be `=` since it will always match and never fall through...

```suggestion
         uri = %URI{scheme: scheme, host: host, port: port, path: path, query: query},
```

mtrudel: Will do manually since I can make the same optimization on the following line as well!

mtrudel:
Good for a 6-10% perf bump on HTTP/1.1 requests!

lib/bandit/http1/handler.ex
```diff
@@ -54,19 +47,73 @@ defmodule Bandit.HTTP1.Handler do
...
+      nil ->
+        case req.version do
+          :"HTTP/1.0" -> {:ok, "", Adapter.get_local_data(req)[:port]}
+          _ -> {:error, "No host header"}
+        end
+
+      error ->
+        error
+    end
+  end
```
ryanwinchester, line 86: Not important at all, but this might be a little more obvious which errors come from where (if I read it correctly)

```suggestion
  defp determine_host_and_port(_request_target, headers, req) do
    case Adapter.get_header(headers, "host") do
      nil ->
        case req.version do
          :"HTTP/1.0" -> {:ok, "", Adapter.get_local_data(req)[:port]}
          _ -> {:error, "No host header"}
        end
        
      host_header ->
        with {:ok, host, port} <- parse_host_header(host_header) do
          case port do
            :undefined -> {:ok, host, Adapter.get_local_data(req)[:port]}
            port -> {:ok, host, port}
          end
        end
    end
  end
```

mtrudel: IIRC I had it laid out that way (or substantially that way) originally, and credo was complaining about the cyclomatic complexity of the function being too high.

mtrudel: I agree that some of these functions can be a bit tricky to read though!

vanderhoop:
Hey, @mtrudel, been enjoying following along with the project since hearing about it in the Phoenix 1.7 rc blog post, and I'm curious about how you identified this particular optimization opportunity, and if there are  any opportunities to optimize URI.parse/1 in the core library as a result of your findings?

mtrudel:
@vanderhoop the original insight came about from talking to @lpil about possible explanations for the discrepancies seen in https://github.com/rawhat/http-benchmarks. It seemed odd that though both Mist and Bandit did fundamentally the same thing, the numbers were quite different. To look deeper into it, I hacked a super simple Hello World server into Bandit, just to try and bisect where the difference may be. Sure enough, this version ended up slightly faster than Mist. A few more bisecting steps later and I'd isolated the difference to Bandit's Plug construction, and here we are.

The task of URI construction is actually one that I'd love to see brought into Plug (prior discussion of this is at https://github.com/elixir-plug/plug/issues/948#issuecomment-833151243). Simply put, constructing a URI from the constituent parts of a raw HTTP request is fraught with choice & policy, and involves at least a few arbitrary decisions that aren't specifically called out in the relevant RFCs. It would be nice to have this normalized between servers by factoring it into the Plug library. It's something I'm thinking about looking into once Bandit gets to its stable 0.8 release stream, likely early 2023.

lpil:
Hey, awesome! Nice one :)

moogle19:
> @moogle19 wanna take a gander?

Sorry, was away for a while and didn't had a chance to check GitHub. 

The PR looks awesome! 

Regarding more performance improvements:
Did you already had a chance to test how much speedup the inlining of specific functions would bring?

mtrudel:
> > @moogle19 wanna take a gander?
> 
> Sorry, was away for a while and didn't had a chance to check GitHub.
> 
> The PR looks awesome!
> 
> Regarding more performance improvements: Did you already had a chance to test how much speedup the inlining of specific functions would bring?

Not specifically, no. I've been iterating on `eprof` output over the past couple of days and I've got about a 15% lift on HTTP/1 requests (I'm going protocol by protocol, just to keep the changes reasonable). 

Most of the work to date is on #61 if you want to take a look / try other stuff! I'm finding the benchmarks provided by the `benchmark` label on PRs to be pretty accurate for runtimes, not so much for memory stats.

PR #59, opened 2022-11-28
Manually run benchmarks

mtrudel:


mtrudel:
<div id="gd"></div>
<script type="module">
    import "https://cdn.plot.ly/plotly-2.16.3.min.js"
    Plotly.newPlot("gd", [{ y: [1, 2, 3] }])
</script>

Issue #63, opened 2022-12-04
Bandit.HTTP1.Adapter.chunk/2 errors on non-string iodata `chunk` argument


rowlandcodes:
**Describe the bug**
The implementation of the `Plug.Conn.chunk` callback in `Bandit.HTTP1.Adapter` uses `byte_size/1` which only accepts `bitstring()`. 

The `Plug.Conn.Adapter` callback `chunk` expects `body :: Conn.body()` for the second argument, where `@type body :: iodata`.

```
** (ArgumentError) errors were found at the given arguments:                                                                                                            
                                                                                                                                                                        
  * 1st argument: not a bitstring                                                                                                                                       
                                                                                                                                                                        
This typically happens when calling Kernel.byte_size/1 with an invalid argument or when performing binary construction or binary concatenation with <> and one of the arguments is not a binary
```

It should be possible to resolve by replacing `byte_size` with `IO.iodata_length` here:
https://github.com/mtrudel/bandit/blob/main/lib/bandit/http1/adapter.ex#L298

I can confirm that this replacement eliminates the error in my codebase.

**To Reproduce**
Pass a list as the second argument for `Plug.Conn.chunk/2` when using `Bandit` as an adapter with HTTP 1.

**Expected behavior**
Chunk sends the iodata without issue.

**Runtime**
 - Elixir version - 1.13.3
 - Erlang version - 25 [erts-13.0.2]
 - Bandit version - 0.6.2
 - Plug version - 1.14.0

**Additional context**

ryanwinchester:
Dang, that's unfortunate. `byte_size/1` is O(1) and `IO.iodata_length` is O(n)....

rowlandcodes:
`IO.iodata_length` inlines to `:erlang.iolist_size`.
It does appear to probably be O(n):
https://github.com/erlang/otp/blob/master/erts/emulator/beam/utils.c#L3020

Though it's obvious that's it's been heavily optimized,
And I couldn't tell you how that code works.

That said, we're already using it, in a sense.
We're giving `Socket.send` an iolist right in the function call.
ThousandIsland `defdelegates` to `:gen_tcp.send`:
https://github.com/mtrudel/thousand_island/blob/main/lib/thousand_island/transports/tcp.ex#L73

`:gen_tcp_socket.send/2` uses `iolist_size` in its implementation:
https://github.com/erlang/otp/blob/master/lib/kernel/src/gen_tcp_socket.erl#L342

[...]

PR #64, opened 2022-12-04
fix: chunk callback expects iodata in second argument

rowlandcodes:
byte_size/1 only accepts bitstrings,
Whereas IO.iodata_length accepts iodata.

Implements fix mentioned in #63


mtrudel:
Looks great! I'll merge this straight away!

Thanks for the PR!

PR #68, opened 2022-12-07
fix: remove unnecessary parens

nbw:
A purely code-cosmetic fix to `port_to_integer`. The parens were not necessary and conflict stylistically with the `port_to_integer` definition above it. 

I feel like I've had this same thing happen to me before when having `mix format` auto execute on save.  

## Note

Looks like my linter also removed some empty spaces. Nice. 🤷


mtrudel:
Good catch! Thanks for the PR!

PR #70, opened 2022-12-12
Factor out a common HTTP pipeline

mtrudel:
Shared concerns for URI building, Conn handling, etc. Also brings improved URI building to h2

### DIFF
--- a/lib/bandit/headers.ex
+++ b/lib/bandit/headers.ex
@@ -1,0 +1,65 @@
+defmodule Bandit.Headers do
+  @moduledoc false
+  # Conveniences for dealing with headers
+
+  # credo:disable-for-this-file Credo.Check.Design.AliasUsage
+
+  def get_header(headers, header) do
+    case List.keyfind(headers, header, 0) do
+      {_, value} -> value
+      nil -> nil
+    end
+  end
+
+  def parse_hostlike_header(host_header) do
+    host_header
+    |> :binary.split(":")
+    |> case do
+      [host, port] ->
+        case Integer.parse(port) do
+          {port, ""} when port > 0 -> {:ok, host, port}
+          _ -> {:error, "Header contains invalid port"}
+        end
+
+      [host] ->
+        {:ok, host, nil}
+    end
+  end
+
+  def get_content_length(headers) do
+    with value when not is_nil(value) <- get_header(headers, "content-length"),
+         {:ok, length} <- parse_content_length(value) do
+      if length >= 0 do
+        {:ok, length}
+      else
+        {:error, "invalid negative content-length header (RFC9110§8.6)"}
+      end
+    else
+      nil -> {:ok, nil}
+      error -> error
+    end
+  end
+
+  defp parse_content_length(value) do
+    case Integer.parse(value) do
+      {length, ""} ->
+        {:ok, length}
+
+      {length, rest} ->
+        rest
+        |> Plug.Conn.Utils.list()
+        |> enforce_unique_value(to_string(length), length)
+
+      :error ->
+        {:error, "invalid content-length header (RFC9112§6.3.5)"}
+    end
+  end
+
+  defp enforce_unique_value([], _str, value), do: {:ok, value}
+
+  defp enforce_unique_value([value | rest], value, int),
+    do: enforce_unique_value(rest, value, int)
+
+  defp enforce_unique_value(_values, _value, _int_value),
+    do: {:error, "invalid content-length header (RFC9112§6.3.5)"}
+end
--- a/lib/bandit/headers.ex
+++ b/lib/bandit/headers.ex
@@ -42,20 +42,16 @@ defmodule Bandit.Headers do
         {:error, "invalid negative content-length header (RFC9110§8.6)"}
 
       {length, rest} ->
-        rest
-        |> Plug.Conn.Utils.list()
-        |> enforce_unique_value(to_string(length), length)
+        if rest |> Plug.Conn.Utils.list() |> all?(to_string(length)),
+          do: {:ok, length},
+          else: {:error, "invalid content-length header (RFC9112§6.3.5)"}
 
       :error ->
         {:error, "invalid content-length header (RFC9112§6.3.5)"}
     end
   end
 
-  defp enforce_unique_value([], _str, value), do: {:ok, value}
-
-  defp enforce_unique_value([value | rest], value, int),
-    do: enforce_unique_value(rest, value, int)
-
-  defp enforce_unique_value(_values, _value, _int_value),
-    do: {:error, "invalid content-length header (RFC9112§6.3.5)"}
+  defp all?([value | rest], value), do: all?(rest, value)
+  defp all?([], _str), do: true
+  defp all?(_values, _value), do: false
 end
--- a/lib/bandit/headers.ex
+++ b/lib/bandit/headers.ex
@@ -27,23 +27,19 @@ defmodule Bandit.Headers do
   end
 
   def get_content_length(headers) do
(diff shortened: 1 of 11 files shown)


lib/bandit/http2/stream.ex
```diff
@@ -100,140 +80,23 @@ defmodule Bandit.HTTP2.Stream do
...
-        :ok
-    end
-  end
-
   defp expected_content_length(headers) do
     case List.keyfind(headers, "content-length", 0) do
-      nil -> nil
-      {_, content_length} -> String.to_integer(content_length)
+      nil -> {:ok, nil}
+      {_, content_length} -> {:ok, String.to_integer(content_length)}
```
ryanwinchester, line 96: Should you do another `case` on `Integer.parse`, or is raising here expected if it doesn't parse?

mtrudel: This is a WiP - I'm going to DRY up @moogle19's recent work on content-length parsing from HTTP/1 and call it here. I wan to make sure that the containing plumbing supports the `{:ok, ...}` wrapping is all.

lib/bandit/headers.ex
```diff
@@ -0,0 +1,65 @@
...
+
+      {length, rest} ->
+        rest
+        |> Plug.Conn.Utils.list()
+        |> enforce_unique_value(to_string(length), length)
+
+      :error ->
+        {:error, "invalid content-length header (RFC9112§6.3.5)"}
+    end
+  end
```
ryanwinchester, line 56: If you put both the nil case handling and the length check and error case in the parse function then both error cases are there and the with can be simplified.

```suggestion
  def get_content_length(headers) do
    with value <- get_header(headers, "content-length") do
      parse_content_length(value)
    end
  end

  defp parse_content_length(nil), do: {:ok, nil}
  
  defp parse_content_length(value) do
    case Integer.parse(value) do
      {length, ""} when length >= 0 ->
        {:ok, length}
        
      {length, ""} ->
        {:error, "invalid negative content-length header (RFC9110§8.6)"}

      {length, rest} ->
        rest
        |> Plug.Conn.Utils.list()
        |> enforce_unique_value(to_string(length), length)

      :error ->
        {:error, "invalid content-length header (RFC9112§6.3.5)"}
    end
  end
  ```

Comment (ryanwinchester) [reply lib/bandit/headers.ex:56]: or should I hold my tongue until it's not a draft? 😅

Comment (ryanwinchester) [reply lib/bandit/headers.ex:56]: Alternatively, if you don't  feel good about a `nil` case for `parse_content_length`

```suggestion
  def get_content_length(headers) do
    case get_header(headers, "content-length") do
      nil -> {:ok, nil}
      value -> parse_content_length(value)
    end
  end

  defp parse_content_length(value) do
    case Integer.parse(value) do
      {length, ""} when length >= 0 ->
        {:ok, length}
        
      {length, ""} ->
        {:error, "invalid negative content-length header (RFC9110§8.6)"}

      {length, rest} ->
        rest
        |> Plug.Conn.Utils.list()
        |> enforce_unique_value(to_string(length), length)

      :error ->
        {:error, "invalid content-length header (RFC9112§6.3.5)"}
    end
  end
  ```

Comment (mtrudel) [reply lib/bandit/headers.ex:56]: This is rapidly becoming 'not a draft' (I have a few more changes but what's up here won't substantially change I don't expect). 

Paging @moogle19 to have an opinion on the content length changes as this is a literal copy and paste of code they added a few weeks ago. Looks fine to me, I think. 

The only thing to consider is that this is one place that speed should win over clarity; it's an incredibly hot path (not that I think that there's anything slow in your changes, just that the guiding light here is a bit different)

Comment (ryanwinchester) [reply lib/bandit/headers.ex:56]: Yeah, I don't expect to see any difference in performance (unless removing the guard clause for `nil` actually helps), but it removes an extra nested conditional and would keep the content-length RFC errors in the parsing function instead of spreading them around.

Comment (mtrudel) [reply lib/bandit/headers.ex:56]: Yeah it's a good change

Comment (ryanwinchester) [reply lib/bandit/headers.ex:56]: I'm not smart enough to contribute features, but I can nitpick the heck out of some code 😅 😬

Comment (mtrudel) [reply lib/bandit/headers.ex]: Went with your second suggestion; it reads great. 

Thanks again!

**Comment by mtrudel** [comment]
Ready for review if any of y'all want to take a look at it. Main changes:

* Factor common 'plug-centric' concerns into `Bandit.Pipeline`. The HTTP/1 and h2 implementations are responsible for
    * Determining the request target & HTTP verb
    * Reading headers
    * Building an adapter concretion of `Plug.Conn.Adapter`
    * Handing all of these off to `Bandit.Pipeline` which takes care of all subsequent tasks in order to send things off to Plug
* Pretty substantial overhaul of the h2 implementation; a lot of setup / header parsing code has moved from `Bandit.HTTP2.Stream` into `Bandit.HTTP2.StreamTask`, and executes within the stream task and not the connection task. This should improve average TTFB latency in heavy load scenarios
* Profiling and tweaking of various HTTP2 code
* Factor common header tasks into `Bandit.Header` (this is a mostly mechanical refactoring)

Feedback / suggestions very welcome! Barring major / breaking feedback, I'm planning on merging this just before the holidays, likely on the 23rd.

**Comment by mtrudel** [comment]
Thoughts on `Bandit.ConnPipeline` in preference to `Bandit.Pipeline`?

File: lib/bandit/headers.ex [hunk 3]
```diff
@@ -0,0 +1,61 @@
...
+  # Conveniences for dealing with headers
+
+  # credo:disable-for-this-file Credo.Check.Design.AliasUsage
+
+  def get_header(headers, header) do
+    case List.keyfind(headers, header, 0) do
+      {_, value} -> value
+      nil -> nil
+    end
+  end
```
ryanwinchester, line 12: You could remove the redundant case on `nil` with `with`. However, without a `@spec` on the function, I do like making the possible returns more obvious.

```suggestion
  def get_header(headers, header) do
    with {_, value} <- List.keyfind(headers, header, 0) do
      value
    end
  end
```

lib/bandit/pipeline.ex
```diff
@@ -0,0 +1,130 @@
...
+
+  @spec run(
+          Plug.Conn.adapter(),
+          transport_info(),
+          Plug.Conn.method(),
+          request_target(),
+          Plug.Conn.headers(),
+          Bandit.plug()
+        ) ::
+          {:ok, Plug.Conn.adapter()} | {:error, term()}
```
ryanwinchester, line 22: It looks like it could also return the websocket info from `maybe_upgrade`. Something more like:
```suggestion
          {:ok, Plug.Conn.adapter()} | {:ok, :websocket, term()} | {:error, term()}
```

lib/bandit/headers.ex
mtrudel: I did exactly that originally, and standardized on making the nil case explicit instead. This is a case where the explicitness of a nil case is worth it, I think.

lib/bandit/pipeline.ex
mtrudel: This isn't the first time that dialyzer has missed a `with` clause like this. Starting to be wary....

lib/bandit/pipeline.ex
```diff
@@ -0,0 +1,130 @@
+defmodule Bandit.Pipeline do
+  @moduledoc false
```
mtrudel, line 2:
```suggestion
  @moduledoc false
  # Provides a common pipeline for HTTP/1.1 and h2 adapters, factoring together shared functionality relating to `Plug.Conn` management
```

PR #76, opened 2023-01-06
Fixup error handling in HTTP/1

### LINKED ISSUE #74: invalid_request, CaseClauseError
I configured Bandit in a Phoenix 1.7-rc.0 / LiveView application. Ever since making the change, I'm getting `invalid_request` errors. In one application, this occurs periodically during deployments, in another one, it happens throughout the day, about 60 times per hour, which makes me think that it has something to do with health checks.

In any case, Cowboy does not log or raise errors under the same circumstances.

Before Bandit 0.6.4, this is what gets logged:

```
message
  GenServer #PID<0.7891.0> terminating
** (stop) :invalid_request
metadata.domain
  otp
metadata.erl_level
  error
metadata.error.initial_call
  
 - 
metadata.error.reason
  {:invalid_request, []}
metadata.error_logger.report_cb
  &:gen_server.format_log/1
metadata.error_logger.tag
  error
severity
  error
```

After upgrading to Bandit 0.6.4, we're getting a `CaseClauseError` instead:

```
CaseClauseError: no case clause matching: {:error, :invalid_request}
  File "lib/thousand_island/handler.ex", line 387, in Bandit.DelegatingHandler.handle_continuation/2
  File "gen_server.erl", line 1123, in :gen_server.try_dispatch/4
  File "gen_server.erl", line 865, in :gen_server.loop/7
  File "proc_lib.erl", line 240, in :proc_lib.init_p_do_apply/3
```

I guess the `:invalid_request` error comes from here: https://github.com/mtrudel/bandit/blob/422651d03b6eb5b2fb10d1ace804880389e7132a/lib/bandit/http1/adapter.ex#L76

This means that `:erlang.decode_packet/3` returns an HTTP error, but unfortunately, Bandit discards the reason.

While we haven't figured out what exactly causes the invalid requests, this should probably be handled gracefully?

mtrudel:
Fixes #74

### DIFF
--- a/lib/bandit/http1/adapter.ex
+++ b/lib/bandit/http1/adapter.ex
@@ -73,8 +73,8 @@ defmodule Bandit.HTTP1.Adapter do
       {:ok, :http_eoh, rest} ->
         {:ok, headers, method, request_target, %{req | state: :headers_read, buffer: rest}}
 
-      {:ok, {:http_error, _reason}, _rest} ->
-        {:error, :invalid_request}
+      {:ok, {:http_error, reason}, _rest} ->
+        {:error, reason}
 
       {:error, reason} ->
         {:error, reason}
--- a/lib/bandit/http1/handler.ex
+++ b/lib/bandit/http1/handler.ex
@@ -26,11 +26,11 @@ defmodule Bandit.HTTP1.Handler do
       else
         {:error, :timeout} ->
           attempt_to_send_fallback(req, 408)
-          {:error, "timeout reading request"}
+          {:error, "timeout reading request", state}
 
         {:error, reason} ->
           attempt_to_send_fallback(req, 400)
-          {:error, reason}
+          {:error, reason, state}
 
         {:ok, :websocket, upgrade_opts} ->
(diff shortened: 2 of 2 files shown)


lib/bandit/http1/adapter.ex
```diff
@@ -73,8 +73,8 @@ defmodule Bandit.HTTP1.Adapter do
       {:ok, :http_eoh, rest} ->
         {:ok, headers, method, request_target, %{req | state: :headers_read, buffer: rest}}
 
-      {:ok, {:http_error, _reason}, _rest} ->
-        {:error, :invalid_request}
+      {:ok, {:http_error, reason}, _rest} ->
+        {:error, reason}
```
woylie, line 77: Since the `reason` is (or can be) a list of integers, just logging the reason is not very helpful, and it is hard to filter/group similar log messages. Can the `{:http_error, reason}` tuple be returned as a whole here, or can the reason be prefixed with `http_error` in other ways otherwise?

mtrudel: Yep. I wrote this before I'd seen the structure of the error (like I said, been trying to repro this for a while). This will most likely change based on the error you'd sent me privately.

PR #78, opened 2023-01-06
Use _bin decode_packet variant

mtrudel:
Trial to see if this is faster


sdball:
For reference if anyone else is curious: https://www.erlang.org/doc/man/erlang.html#decode_packet-3

> The variants `http_bin` and `httph_bin` return strings (HttpString) as binaries instead of lists.

mtrudel:
This is one of the hottest paths in the entire library, and the previous version dated from the original `0.1.0` experiment phase version of Bandit. It pays to go over old code from time to time!

PR #80, opened 2023-01-07
Add Bandit.IO

Schultzer:
This PR adds a new Bandit.IO module.

I’ve not been able to disasmble (https://github.com/scout119/beamdasm is giving me errors with VSCodium) the code to inspect whats really going on and we might just be gaming the benchmark.

So even thought we’re winning on synthetic benchmark and tests then I recommend testing and benchmarking it against real world data and keep in mind that this might be faster then binary concat on big IOdata and slower on small.

This leaves us with the last question I can think of: Why is this faster?

Please bare with me as I’m not an expert on the BEAM so if anyone reading this with some insights then I like to hear them.

Anyways I believe it’s due to the code having less branches, instead of leveraging deeply nested if else statements then we use pattern matching and functions calls.

There is also for some reason a big reduction in memory usage, which I can’t really explain why.

In the end I hope this leads us in the right direction, which might be to have better IO primitives in the BEAM.

```elixir
➜  iolist_test-master iex -S mix
Erlang/OTP 25 [erts-13.1.2] [source] [64-bit] [smp:10:10] [ds:10:10:10] [async-threads:1] [jit]

Compiling 2 files (.ex)
Interactive Elixir (1.14.1) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)> IOListTest.Benchmark.run
Operating System: macOS
CPU Information: Apple M1 Max
Number of Available Cores: 10
Available memory: 64 GB
Elixir 1.14.1
Erlang 25.1.2

warmup: 2 s
time: 10 s
memory time: 2 s
parallel: 1
inputs: none specified
Estimated total run time: 1.40 min

Benchmarking binary_concat...
Benchmarking cow_lib_split...
Benchmarking iodata_split...
Benchmarking iodata_split_3...
Benchmarking iodata_split_4...
Benchmarking iodata_split_bandit...

Name                          ips        average  deviation         median         99th %
iodata_split_4            6327.92       0.158 ms   ±966.07%      0.0600 ms      0.0772 ms
iodata_split_bandit       5253.82       0.190 ms   ±770.61%       0.101 ms       0.119 ms
binary_concat               13.10       76.33 ms    ±16.12%       79.82 ms      123.63 ms
iodata_split_3               4.74      210.84 ms    ±39.60%      260.76 ms      270.39 ms
iodata_split                 4.42      226.04 ms    ±42.84%      286.76 ms      296.83 ms
cow_lib_split                2.95      338.55 ms    ±48.59%      439.81 ms      459.11 ms

Comparison:
iodata_split_4            6327.92
iodata_split_bandit       5253.82 - 1.20x slower +0.0323 ms
binary_concat               13.10 - 483.01x slower +76.17 ms
iodata_split_3               4.74 - 1334.18x slower +210.68 ms
iodata_split                 4.42 - 1430.33x slower +225.88 ms
cow_lib_split                2.95 - 2142.33x slower +338.39 ms

Memory usage statistics:

Name                   Memory usage
iodata_split_4             0.144 MB
iodata_split_bandit        0.130 MB - 0.91x memory usage -0.01316 MB
binary_concat              38.20 MB - 266.11x memory usage +38.05 MB
iodata_split_3             75.32 MB - 524.71x memory usage +75.17 MB
iodata_split               75.32 MB - 524.71x memory usage +75.17 MB
cow_lib_split             115.00 MB - 801.14x memory usage +114.85 MB

:ok
iex(2)>
```


lib/bandit/io.ex
```diff
@@ -0,0 +1,55 @@
...
+  remainder of the iolist beyond it as a tuple. Returns `{:error, :length_exceeded}` if there are
+  not `length` bytes remaining in the iolist.
+  """
+  @spec split_iodata(iodata(), pos_integer(), iodata(), iodata()) :: {iodata(), iodata()} | {:error, :length_exceeded}
+  def split_iodata(iodata, length, head \\ [], acc \\ [])
+  def split_iodata(iodata, 0, [] = head, []), do: {head, iodata}
+  def split_iodata([], _length, [], []), do: {:error, :length_exceeded}
+  def split_iodata("", _length, [], []), do: {:error, :length_exceeded}
+  def split_iodata([byte | _] = iodata, 0, head, []) when is_integer(byte), do: {head, iodata}
+  def split_iodata(_iodata, 0, head, [] = acc), do: {head, acc}
```
moogle19, line 15: This should return the iodata as second element and not the empty `acc`.
This also explains why the benchmark was that fast, since it dismissed a part of the data.

lib/bandit/io.ex
```diff
@@ -0,0 +1,55 @@
...
+  def split_iodata(iodata, 0, head, acc), do: {head, [iodata | acc]}
+  def split_iodata([], _length, head, acc), do: {head, acc}
+  def split_iodata([[] | iodata], length, head, acc), do: split_iodata(iodata, length, head, acc)
+  def split_iodata(["" | iodata], length, head, acc), do: split_iodata(iodata, length, head, acc)
+  def split_iodata([byte | iodata], length, head, acc) when is_integer(byte), do: split_iodata(iodata, length - 1, head ++ [byte], acc)
+  def split_iodata([h | iodata], length, head, [] = acc) when h != [] do
```
ryanwinchester, line 21: Is pattern-matching preferable over guards when you can do it? (Or does it not matter?)

```suggestion
  def split_iodata([[_ | _] = h | iodata], length, head, acc) do
```

lib/bandit/io.ex
```diff
@@ -0,0 +1,55 @@
...
+    head_length = IO.iodata_length(h)
+    if head_length <= length do
+      split_iodata(iodata, length - head_length, [head | h], acc)
+    else
+      split_iodata(h, length, head, iodata)
+    end
+  end
+  def split_iodata([h | iodata], length, head, acc) when h != [] do
```
ryanwinchester, line 29: Is pattern-matching preferable over guards when you can do it? (Or does it not matter?)

```suggestion
  def split_iodata([[_ | _] = h | iodata], length, head, acc) do
```

lib/bandit/io.ex
Schultzer: Thanks for the review, that could explain the reduction in memory.

lib/bandit/io.ex
Schultzer: Good question, I'm not really sure but it wouldn't surprise me if it already got optimized away.

lib/bandit/io.ex
Schultzer: Good question, I'm not really sure but it wouldn't surprise me if it already got optimized away.

lib/bandit/io.ex
```diff
@@ -0,0 +1,55 @@
...
+      ## so it could indicate that it's redudent.
+      split_iodata(iodata, length - head_length, [head | h], acc)
+    else
+      split_iodata(h, length, head, [iodata | acc])
+    end
+  end
+  def split_iodata(<<iodata::binary()>>, length, head, acc) do
+    case iodata do
+      <<h::binary-size(length), rest::binary>> ->
+        {[head | h], [rest | acc]}
```
moogle19, line 43: The `h` should be wrapped in a list here (like `[head | [h]]`) to avoid creating an improper list (which also makes it faster in my benchmarks).

lib/bandit/io.ex
```diff
@@ -0,0 +1,55 @@
...
+
+      iodata ->
+        {[head | iodata], acc}
```
moogle19, line 46: Same as above for `iodata` here.

lib/bandit/io.ex
mtrudel: iolists are specifically allowed to be improper, though if it's faster that's rad!

lib/bandit/io.ex
mtrudel: Minor preference for pattern matching here just for legibility.

mtrudel:
I updated a couple of low hanging fruit calls in the HTTP2 stack with this function and set the benchmarker on it. Let's see if it boils!

mtrudel:
Update: no big wins on the benchmarks, which isn't surprising since the cases we're optimizing aren't ones that the benchmark covers. Let me look for more opportunities to apply this.

Schultzer:
> Update: no big wins on the benchmarks, which isn't surprising since the cases we're optimizing aren't ones that the benchmark covers. Let me look for more opportunities to apply this.

Thanks for benchmarking this I'm gonna close this for now since it isn't bringing any proformance gains.

mtrudel:
On the contrary I think the places that this touches are going to see huge gains; it's just that they're not cases the benchmark currently covers.

Issue #93, opened 2023-02-06
Missing common HTTP status codes


jclem:
I have noticed that there are common HTTP status codes that Bandit is not aware of. For example, the very common [422 Unprocessable Entity](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422) returns "Unknown Status Code" from Bandit, but there are many others.

Is this a conscious choice, or would you accept a pull request adding some common, well-documented HTTP status codes?

At a minimum, I think it makes sense to add most of [these codes](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status) from RFC 9110 and other places.

moogle19:
Couldn't we use `Plug.Conn.Status.reason_phrase/1` here and drop the `reason_for_status/1` completely?

jclem:
@moogle19's suggestion may be a better option than the PR we're discussing @mtrudel.

See the status codes as of Plug 1.14.0 [here](https://github.com/elixir-plug/plug/blob/d7940552e365e0d3245bb78ca99429ea5c324f7d/lib/plug/conn/status.ex#L8-L72) (although note that Plug, too, does include "418 I'm a teapot" 😆). It is identical to the list after my PR, but also includes "306 Switch Proxy" (my assumption is that Plug includes this now removed status code in this reserved space to handle servers talking to very old clients that need HTTP 306).

I'll open a separate PR so that we can compare them.

PR #94, opened 2023-02-06
Add some common HTTP status codes

jclem:
This adds a few common HTTP status codes, mostly from the WebDAV specification. Among them are very common status codes such as "422 Unprocessable Entity" and "429 Too Many Requests". My primary source was [this page from MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status). It also corrects the capitalization of "505 HTTP Version Not Supported" to match [RFC 9110](https://datatracker.ietf.org/doc/html/rfc9110#name-505-http-version-not-suppor).

Hopefully, none of the added codes are controversial—the list now also matches the [status code list in the Go language](https://github.com/golang/go/blob/bf8d142b4ebe9f84e75b20a87d292d2d6ec13fab/src/net/http/status.go#L9-L77).

Closes https://github.com/mtrudel/bandit/issues/93


lib/bandit/http1/adapter.ex
```diff
@@ -335,20 +335,26 @@ defmodule Bandit.HTTP1.Adapter do
   response_reasons = %{
     100 => "Continue",
     101 => "Switching Protocols",
+    102 => "Processing",
+    103 => "Early Hints",
```
mtrudel, line 339: Not in 9110

lib/bandit/http1/adapter.ex
```diff
@@ -335,20 +335,26 @@ defmodule Bandit.HTTP1.Adapter do
...
     200 => "OK",
     201 => "Created",
     202 => "Accepted",
     203 => "Non-Authoritative Information",
     204 => "No Content",
     205 => "Reset Content",
     206 => "Partial Content",
+    207 => "Multi-Status",
+    208 => "Already Reported",
+    226 => "IM Used",
```
mtrudel, line 349: Not in 9110

lib/bandit/http1/adapter.ex
```diff
@@ -367,12 +373,28 @@ defmodule Bandit.HTTP1.Adapter do
     415 => "Unsupported Media Type",
     416 => "Requested range not satisfiable",
     417 => "Expectation Failed",
+    418 => "I'm a teapot",
```
mtrudel, line 376: Not in 9110 (It IS cute though)

lib/bandit/http1/adapter.ex
```diff
@@ -367,12 +373,28 @@ defmodule Bandit.HTTP1.Adapter do
...
+    421 => "Misdirected Request",
+    422 => "Unprocessable Entity",
+    423 => "Locked",
+    424 => "Failed Dependency",
+    425 => "Too Early",
```
mtrudel, line 381: Not in 9110

lib/bandit/http1/adapter.ex
```diff
@@ -367,12 +373,28 @@ defmodule Bandit.HTTP1.Adapter do
...
+    426 => "Upgrade Required",
+    428 => "Precondition Required",
+    429 => "Too Many Requests",
+    431 => "Request Header Fields Too Large",
+    451 => "Unavailable For Legal Reasons",
```
mtrudel, line 386: Not in 9110

lib/bandit/http1/adapter.ex
```diff
@@ -367,12 +373,28 @@ defmodule Bandit.HTTP1.Adapter do
...
     502 => "Bad Gateway",
     503 => "Service Unavailable",
     504 => "Gateway Time-out",
-    505 => "HTTP Version not supported"
+    505 => "HTTP Version Not Supported",
+    506 => "Variant Also Negotiates",
+    507 => "Insufficient Storage",
+    508 => "Loop Detected",
+    510 => "Not Extended",
+    511 => "Network Authentication Required"
```
mtrudel, line 397: Not in 9110

mtrudel:
I can get behind (most of) this! A couple of things before committing to this, to make sure we're closing this off properly and comprehensively:

* I'll try and remember where I sourced this list from (I know I didn't write it myself, but rather adapted it nearly verbatim from another project). I certainly didn't omit any of these codes on purpose, so I'd like to try and remember where the original list came from
* I've always built Bandit based directly on relevant RFCs and not second-hand docs such as MDN (not that there's anything wrong with them, but being dogmatic about RFC compliance makes it easier to avoid unintentional cargo-culting IMO). As you say, RFC9110§15 is the authority here, and there are a number of codes in this PR which aren't in RFC9110. Even though they may be in use (either widely or not) I'd prefer to keep to the letter of the RFC.

mtrudel (changes requested):

jclem:
None of these are sourced second-hand, they're all from RFCs, including "418 I'm a teapot" which is from [RFC 2324](https://www.rfc-editor.org/rfc/rfc2324#section-2.3.2), which [may be reserved](https://lists.w3.org/Archives/Public/ietf-http-wg/2017JulSep/0332.html). I did not include it here for cuteness—just completeness. It is included in Go, Python, and many other language standard libraries.

The other status codes here are mostly from WebDAV. I considered including each RFC here and can do so if you wish. The related RFCs for each of these are all actually in that Go link.

lib/bandit/http1/adapter.ex
jclem: [RFC 6585](https://www.rfc-editor.org/rfc/rfc6585#section-6)

lib/bandit/http1/adapter.ex
jclem: [RFC 7725](https://datatracker.ietf.org/doc/html/rfc7725#section-3)

Proposed, but common.

lib/bandit/http1/adapter.ex
jclem: [RFC 8470](https://datatracker.ietf.org/doc/html/rfc8470#section-5.2)

lib/bandit/http1/adapter.ex
jclem: [RFC 2324](https://www.rfc-editor.org/rfc/rfc2324#section-2.3.2)

See many language stdlibs, including [Python](https://github.com/python/cpython/blob/7a253103d4c64fcca4c0939a584c2028d8da6829/Lib/http/__init__.py#L129-L130) (and Go, linked in the PR body).

lib/bandit/http1/adapter.ex
jclem: [RFC 3229](https://www.rfc-editor.org/rfc/rfc3229#section-10.4.1)

lib/bandit/http1/adapter.ex
jclem: [RFC 8297](https://httpwg.org/specs/rfc8297.html)

It would be odd to serve early hints from Bandit apps that serve HTML and not have this text included.

jclem:
See https://github.com/mtrudel/bandit/pull/96 for a potentially better alternative.

PR #96, opened 2023-02-07
Use Plug.Conn.Status.reason_phrase/1

### LINKED ISSUE #93: Missing common HTTP status codes

jclem:
As an alternative to [Add some common HTTP status codes #94](https://github.com/mtrudel/bandit/pull/94), this PR removes Bandit's logic for generating status text from HTTP status codes and replaces it with [`Plug.Conn.Status.reason_phrase/1`](https://github.com/elixir-plug/plug/blob/d7940552e365e0d3245bb78ca99429ea5c324f7d/lib/plug/conn/status.ex#L146-L174).

Thanks to @moogle19 for [suggesting this](https://github.com/mtrudel/bandit/issues/93#issuecomment-1420385066).

The [status code list from Plug](https://github.com/elixir-plug/plug/blob/d7940552e365e0d3245bb78ca99429ea5c324f7d/lib/plug/conn/status.ex#L8-L72) matches the existing Bandit list, plus what was added in #94, plus HTTP 306 Switch Proxy, a deprecated status code in reserved space that some servers may return when presumably communicating with very old clients.

Currently, this PR is a breaking change to Bandit, because Plug's default behavior is to [raise when an unknown status code is encountered](https://github.com/elixir-plug/plug/blob/d7940552e365e0d3245bb78ca99429ea5c324f7d/lib/plug/conn/status.ex#L153-L174). However, non-standard HTTP status codes can still be registered via configuration: `config :plug, :statuses, %{998 => "Not An RFC Status Code"}`.

Overall, I think this is an improvement to Bandit and favorable over #94:

- Removes a divergence between Bandit and Plug.
- Provides an escape hatch for non-standard HTTP status codes.
- Removes the burden on Bandit maintainers of having to decide which status codes to include.

If we wanted to land this change in a non-breaking manner (assuming that the change from "Unknown Status Code" to the newly included status texts is not breaking), we could instead apply this patch to this PR and land the breaking version later:

```patch
diff --git c/lib/bandit/http1/adapter.ex w/lib/bandit/http1/adapter.ex
index 1ea4525..3f1ba2c 100644
--- c/lib/bandit/http1/adapter.ex
+++ w/lib/bandit/http1/adapter.ex
@@ -1,6 +1,8 @@
 defmodule Bandit.HTTP1.Adapter do
   @moduledoc false
 
+  require Logger
+
   @type state :: :new | :headers_read | :no_body | :body_read | :sent | :chunking_out
 
   @behaviour Plug.Conn.Adapter
@@ -302,7 +304,7 @@ defmodule Bandit.HTTP1.Adapter do
       " ",
       to_string(status),
       " ",
-      Plug.Conn.Status.reason_phrase(status),
+      reason_phrase(status),
       "\r\n",
       Enum.map(headers, fn {k, v} -> [k, ": ", v, "\r\n"] end),
       "\r\n"
@@ -331,4 +333,15 @@ defmodule Bandit.HTTP1.Adapter do
   def get_http_protocol(%__MODULE__{version: version}), do: version
 
   def keepalive?(%__MODULE__{keepalive: keepalive}), do: keepalive
+
+  defp reason_phrase(status) do
+    Plug.Conn.Status.reason_phrase(status)
+  rescue
+    e in ArgumentError ->
+      # Raising ArgumentError would be breaking, so we rescue until the next
+      # major version. Log the error as a debug message so users are notified
+      # they can provide configuration for non-standard codes.
+      Logger.debug(e.message)
+      "Unknown Status Code"
+  end
 end
```

Closes #93

### DIFF
--- a/lib/bandit/http1/adapter.ex
+++ b/lib/bandit/http1/adapter.ex
@@ -302,7 +302,7 @@ defmodule Bandit.HTTP1.Adapter do
       " ",
       to_string(status),
       " ",
-      reason_for_status(status),
+      Plug.Conn.Status.reason_phrase(status),
       "\r\n",
       Enum.map(headers, fn {k, v} -> [k, ": ", v, "\r\n"] end),
       "\r\n"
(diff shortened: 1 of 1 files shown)


mtrudel:
Yeah I love this approach (and honestly, I'm vexed as to why I didn't go this route originally; I certainly knew about this). Great thinking @moogle19 !

I'm fine with the breaking change coming in now - we don't have test coverage one way or the other & it's a pretty tight corner case regardless (heck, even teapots won't cause it! :)

Are y'all OK with this as it stands now? I am...

jclem:
Since Bandit is still pre-1.0, it seems fair to introduce this as a breaking change. I'd feel differently if 1.0 had been published.

I'm 👍 landing this as-is.

mtrudel:
> Since Bandit is still pre-1.0, it seems fair to introduce this as a breaking change. I'd feel differently if 1.0 had been published.

Yup.

> I'm 👍 landing this as-is.

Merging!

mtrudel:
Thanks for the PR!

PR #97, opened 2023-02-17
Fix host header consumption for ipv6 hostnames

derekkraan:
Good morning!

Last time I found something, I missed my chance to get a PR in, so I'm hoping to rectify that this time ;)

I was seeing errors when accessing my local server using its IPv6 address as hostname (some HTTP clients unfortunately convert localhost to `[::1]` automatically). So I did some digging and discovered that bandit just didn't cover this case at all.

I think this PR should just about cover it. Let me know if you'd prefer to see something changed in here.

Also I have to say, the more I read the code the more impressed I am with bandit / thousand island. It isn't a trivial task to make code that is this clear, let alone an http server implementation. I was able to find and fix this (including puzzling over how to make the actual function body not look insane) in about 40 minutes.

### DIFF
--- a/lib/bandit/headers.ex
+++ b/lib/bandit/headers.ex
@@ -11,6 +11,22 @@ defmodule Bandit.Headers do
     end
   end
 
+  # covers ipv6 addresses, which look like this: `[::1]:4000` as defined in RFC2732
+  def parse_hostlike_header("[" <> _ = host_header) do
+    host_header
+    |> :binary.split("]:")
+    |> case do
+      [host, port] ->
+        case Integer.parse(port) do
+          {port, ""} when port > 0 -> {:ok, host <> "]", port}
+          _ -> {:error, "Header contains invalid port"}
+        end
+
+      [host] ->
+        {:ok, host, nil}
+    end
+  end
+
   def parse_hostlike_header(host_header) do
     host_header
     |> :binary.split(":")
--- a/test/bandit/http1/request_test.exs
+++ b/test/bandit/http1/request_test.exs
@@ -186,6 +186,7 @@ defmodule HTTP1RequestTest do
       assert Jason.decode!(body)["host"] == "banana"
     end
 
+    @tag :skip
     test "derives ipv6 host from the URI, even if it differs from host header", context do
       client = SimpleHTTP1Client.tcp_client(context)
 
--- a/test/bandit/http1/request_test.exs
+++ b/test/bandit/http1/request_test.exs
@@ -61,6 +61,17 @@ defmodule HTTP1RequestTest do
       assert Jason.decode!(body)["port"] == 1234
     end
 
+    test "derives host from host header with ipv6 host", context do
+      client = SimpleHTTP1Client.tcp_client(context)
+
+      SimpleHTTP1Client.send(client, "GET", "/echo_components", [
+        "host: [FEDC:BA98:7654:3210:FEDC:BA98:7654:3210]"
+      ])
+
+      assert {:ok, "200 OK", _headers, body} = SimpleHTTP1Client.recv_reply(client)
+      assert Jason.decode!(body)["host"] == "[FEDC:BA98:7654:3210:FEDC:BA98:7654:3210]"
+    end
+
     test "derives host and port from host header with ipv6 host", context do
       client = SimpleHTTP1Client.tcp_client(context)
       SimpleHTTP1Client.send(client, "GET", "/echo_components", ["host: [::1]:1234"])
@@ -173,6 +184,20 @@ defmodule HTTP1RequestTest do
       SimpleHTTP1Client.send(client, "GET", "http://banana/echo_components", ["host: orange"])
       assert {:ok, "200 OK", _headers, body} = SimpleHTTP1Client.recv_reply(client)
       assert Jason.decode!(body)["host"] == "banana"
+    end
+
+    test "derives ipv6 host from the URI, even if it differs from host header", context do
+      client = SimpleHTTP1Client.tcp_client(context)
+
+      SimpleHTTP1Client.send(
+        client,
+        "GET",
+        "http://[FEDC:BA98:7654:3210:FEDC:BA98:7654:3210]/echo_components",
+        ["host: orange"]
+      )
+
+      assert {:ok, "200 OK", _headers, body} = SimpleHTTP1Client.recv_reply(client)
+      assert Jason.decode!(body)["host"] == "[FEDC:BA98:7654:3210:FEDC:BA98:7654:3210]"
     end
 
     @tag capture_log: true
--- a/test/bandit/http1/request_test.exs
+++ b/test/bandit/http1/request_test.exs
@@ -61,6 +61,14 @@ defmodule HTTP1RequestTest do
       assert Jason.decode!(body)["port"] == 1234
     end
 
+    test "derives host and port from host header with ipv6 host", context do
+      client = SimpleHTTP1Client.tcp_client(context)
+      SimpleHTTP1Client.send(client, "GET", "/echo_components", ["host: [::1]:1234"])
+      assert {:ok, "200 OK", _headers, body} = SimpleHTTP1Client.recv_reply(client)
+      assert Jason.decode!(body)["host"] == "[::1]"
+      assert Jason.decode!(body)["port"] == 1234
+    end
+
     @tag capture_log: true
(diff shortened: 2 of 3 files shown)


lib/bandit/headers.ex
```diff
@@ -11,6 +11,22 @@ defmodule Bandit.Headers do
...
+    |> :binary.split("]:")
+    |> case do
+      [host] ->
+        {:ok, host, nil}
+
+      [host, port] ->
+        case Integer.parse(port) do
+          {port, ""} when port > 0 -> {:ok, host <> "]", port}
+          _ -> {:error, "Header contains invalid port"}
+        end
```
mtrudel, line 26: Suggest to reorder these clauses so they match those in the clause just below:

```suggestion
      [host, port] ->
        case Integer.parse(port) do
          {port, ""} when port > 0 -> {:ok, host <> "]", port}
          _ -> {:error, "Header contains invalid port"}
        end
        
      [host] ->
        {:ok, host, nil}        
```

lib/bandit/headers.ex
```diff
@@ -11,6 +11,22 @@ defmodule Bandit.Headers do
     end
   end
 
+  # covers ipv6 addresses, which look like this: `[::1]:4000`
```
mtrudel, line 14:
```suggestion
  # covers ipv6 addresses, which look like this: `[::1]:4000` as defined in RFC2732
```

mtrudel (changes requested):
A couple of stylistic / doc changes suggested. Also please see the top-level comment on the PR regarding complete coverage of the cases in RFC2732

mtrudel:
Interesting! This is definitely a thing we should support (and to be honest, isn't something I'd thought about before). I've left a couple of small style suggestions in a review, but there are two slightly larger issues we should tie up here as well.

* This should ideally provide complete coverage of all RFC 2732 cases. It doesn't look like this version would cover the case where the Host field is just an IPv6 with no trailing port (ie: `Host: [123:456::1]`). We should cover that case as well (it's a pretty easy change I think)

* `parse_hostlike_header` is also called when parsing an absolute request line. We should cover those cases in tests as well. See the tests in the https://github.com/mtrudel/bandit/blob/main/test/bandit/http1/request_test.exs#L148 stanza for some examples. Note that I don't think there's any code changes required here; I just think we should get those cases under test as well

derekkraan:
I've made those changes you requested and force pushed.

Re your top-level comments:
- I believe it should cover the case when there is no trailing port number, that should match `[host] = :binary.split("]:")` if I'm correct. What I did notice, because I did run the test without the port suffix (ie just with `host: [::1]`), is that bandit seems to return a random port number in this case. I'm not sure at all what's going on here to be honest, it's not coming from this part of the code.
- I can add the two cases (ipv6 with and without port) to this section of the tests as well.

derekkraan:
Turns out there _are_ code changes required to support ipv6 IPs in the URI.

derekkraan:
It looks like `:erlang.decode_packet/3` isn't able to handle the IPv6 address in the URI.

derekkraan:
I pushed another change so you can see what I'm talking about.

mtrudel:
It's pretty clear to me that the same syntax should. be supported in request lines (RFC9112§3.1.2 references RFC3986§4.3. RFC3986 obsoletes RFC2732 but includes the same bracket-delimited clause). This looks squarely like a bug in decode_packet. I'd say we add the relevant tests even if they fail and mark them with `@tag :skip` until / if they're fixed upstream.

derekkraan:
Sounds like a plan. I'll take another look at this on Monday when I have a clear head again. Have a good weekend!

derekkraan:
For the paper trail: https://github.com/erlang/otp/blob/master/erts/emulator/beam/packet_parser.c#L556-L586 appears to be in close proximity to the location where erlang is saying "this part of URL goes until the `:`".

mtrudel:
Looks great! I'll just do a quick copy of those tests over to HTTP/2 for parity and this is good to go!

Thanks for the PR!

mtrudel:
I've got a couple of changes that I can't seem to push up to GitHub (I suspect you probably have your permissions locked down or possibly forgot to allow maintainer edits when creating this PR). Could you please apply the following on this branch?

```diff
diff --git a/test/bandit/http1/request_test.exs b/test/bandit/http1/request_test.exs
index a23092b..3ec677f 100644
--- a/test/bandit/http1/request_test.exs
+++ b/test/bandit/http1/request_test.exs
@@ -186,6 +186,7 @@ defmodule HTTP1RequestTest do
       assert Jason.decode!(body)["host"] == "banana"
     end
 
+    # Skip this test since there is a bug in :erlang.decode_packet. See https://github.com/mtrudel/bandit/pull/97
     @tag :skip
     test "derives ipv6 host from the URI, even if it differs from host header", context do
       client = SimpleHTTP1Client.tcp_client(context)
diff --git a/test/bandit/http2/protocol_test.exs b/test/bandit/http2/protocol_test.exs
index 024f824..8c7b257 100644
--- a/test/bandit/http2/protocol_test.exs
+++ b/test/bandit/http2/protocol_test.exs
@@ -1449,6 +1449,41 @@ defmodule HTTP2ProtocolTest do
       assert Jason.decode!(body)["port"] == 1234
     end
 
+    test "derives host from host header with ipv6 host", context do
+      socket = SimpleH2Client.setup_connection(context)
+
+      headers = [
+        {":method", "GET"},
+        {":path", "/echo_components"},
+        {":scheme", "https"},
+        {"host", "[FEDC:BA98:7654:3210:FEDC:BA98:7654:3210]"}
+      ]
+
+      SimpleH2Client.send_headers(socket, 1, true, headers)
+
+      assert SimpleH2Client.successful_response?(socket, 1, false)
+      {:ok, 1, true, body} = SimpleH2Client.recv_body(socket)
+      assert Jason.decode!(body)["host"] == "[FEDC:BA98:7654:3210:FEDC:BA98:7654:3210]"
+    end
+
+    test "derives host and port from host header with ipv6 host", context do
+      socket = SimpleH2Client.setup_connection(context)
+
+      headers = [
+        {":method", "GET"},
+        {":path", "/echo_components"},
+        {":scheme", "https"},
+        {"host", "[::1]:1234"}
+      ]
+
+      SimpleH2Client.send_headers(socket, 1, true, headers)
+
+      assert SimpleH2Client.successful_response?(socket, 1, false)
+      {:ok, 1, true, body} = SimpleH2Client.recv_body(socket)
+      assert Jason.decode!(body)["host"] == "[::1]"
+      assert Jason.decode!(body)["port"] == 1234
+    end
+
     @tag capture_log: true
     test "resets stream if port cannot be parsed from host header", context do
       socket = SimpleH2Client.setup_connection(context)
@@ -1661,6 +1696,24 @@ defmodule HTTP2ProtocolTest do
       assert Jason.decode!(body)["host"] == "banana"
     end
 
+    test "derives ipv6 host from the URI, even if it differs from host header", context do
+      socket = SimpleH2Client.setup_connection(context)
+
+      headers = [
+        {":method", "GET"},
+        {":path", "/echo_components"},
+        {":scheme", "https"},
+        {":authority", "[FEDC:BA98:7654:3210:FEDC:BA98:7654:3210]:#{context.port}"},
+        {"host", "orange"}
+      ]
+
+      SimpleH2Client.send_headers(socket, 1, true, headers)
+
+      assert SimpleH2Client.successful_response?(socket, 1, false)
+      {:ok, 1, true, body} = SimpleH2Client.recv_body(socket)
+      assert Jason.decode!(body)["host"] == "[FEDC:BA98:7654:3210:FEDC:BA98:7654:3210]"
+    end
+
     test "derives port from host header, even if it differs from host header", context do
       socket = SimpleH2Client.setup_connection(context)
 
```

mtrudel (approved):

derekkraan:
@mtrudel I made those changes you asked for.

Not sure why it didn't allow you to push changes, "Allow edits by maintainers" is checked. Maybe because I did the lazy thing and just pushed my commits to the `main` branch in my fork?

derekkraan:
Put in a PR to fix this in erlang. https://github.com/erlang/otp/pull/6900

mtrudel:
This looks great!

derekkraan:
Included in OTP 26: https://erlang.org/download/OTP-26.0.README (search for `OTP-18540`)

mtrudel:
Added a note to un-skip the relevant test once we're OTP26+ (not that that'll be anytime soon!)

Thanks for the update!

PR #99, opened 2023-02-21
Support extended connection close return values

mtrudel:
Implements Bandit support for close tuples as described in https://github.com/phoenixframework/websock_adapter/issues/2; supports callbacks to return {:stop, reason, code, state} or {:stop, reason, {code, detail}, state}.

Pairs with / requires https://github.com/phoenixframework/websock/pull/6.

PR #103, opened 2023-02-28
Add support for custom close code and reason

### LINKED ISSUE #102: Error when using graphql-ws protocol
I'm trying to use Bandit with an application that uses the graphql-ws protocol:
https://github.com/wodup/absinthe_graphql_ws

The protocol allows for sending connection params (e.g. a token), and the server can close the socket if e.g. the token is invalid:
https://github.com/wodup/absinthe_graphql_ws/blob/main/lib/absinthe/graphql_ws/transport.ex#L129

However I get this error when using Bandit. With a regular Phoenix/cowboy setup, this error does not occur.
```
** (FunctionClauseError) no function clause matching in Bandit.WebSocket.Connection.do_deflate/3
    (bandit 0.6.9) lib/bandit/websocket/connection.ex:236: Bandit.WebSocket.Connection.do_deflate({:close, 4403, "Forbidden"}, %ThousandIsland.Socket{socket: #Port<0.139>, transport_module: ThousandIsland.Transports.TCP, read_timeout: 60000, span: %ThousandIsland.Telemetry{span_name: :connection, span_id: "XJGNI6FEZC4RZZ22", start_time: -576460747475728060}}, %Bandit.WebSocket.Connection{websock: WodUpWeb.GraphqlWSSocket, websock_state: %Absinthe.GraphqlWS.Socket{absinthe: %{opts: [context: %{pubsub: WodUpWeb.Endpoint}], pipeline: {Absinthe.GraphqlWS.Socket, :absinthe_pipeline}, schema: WodUpWeb.Graphql.Schema}, connect_info: %{}, endpoint: WodUpWeb.Endpoint, handler: WodUpWeb.GraphqlWSSocket, keepalive: 30000, pubsub: WodUp.PubSub, assigns: %{}, initialized?: false, subscriptions: %{}}, state: :open, compress: nil, fragment_frame: nil, span: %Bandit.Telemetry{span_name: :websocket, span_id: "VO6L3P3TOR4NWFO5", start_time: -576460746978892054}, metrics: %{recv_text_frame_bytes: 131, recv_text_frame_count: 1}})
    (bandit 0.6.9) lib/bandit/websocket/handler.ex:42: anonymous fn/3 in Bandit.WebSocket.Handler.handle_data/3
```

Here's the relevant code from the reference JS implementation of graphql-ws if that helps:
https://github.com/enisdenjo/graphql-ws/blob/799cfc7bbe0f6d1a5c90b4880f02c58c5c3a06d4/src/server.ts#L612-L614

https://github.com/enisdenjo/graphql-ws/blob/799cfc7bbe0f6d1a5c90b4880f02c58c5c3a06d4/src/__tests__/server.ts#L437-L458

moogle19:
This fixes #102.

Currently we don't support closing WebSocket connections with a custom code and/or message cleanly (Only `normal` closing without a message or an `error` which only changes the code are possible).

This adds support for sending `close` frames similar to cowboy, by passing either `:error`, `{:error, message}`  or `{:error, code, message}` tuples in the reply tuple.

### DIFF
--- a/lib/bandit/websocket/connection.ex
+++ b/lib/bandit/websocket/connection.ex
@@ -223,15 +223,6 @@ defmodule Bandit.WebSocket.Connection do
       {:ok, websock_state} ->
         {:continue, %{connection | websock_state: websock_state}}
 
-      {:reply, _status, :close, websock_state} ->
-        do_close(@normal_close_code, socket, %{connection | websock_state: websock_state})
-
-      {:reply, _status, {:close, msg}, websock_state} ->
-        do_close(@normal_close_code, msg, socket, %{connection | websock_state: websock_state})
-
-      {:reply, _status, {:close, code, msg}, websock_state} ->
-        do_close(code, msg, socket, %{connection | websock_state: websock_state})
-
       {:reply, _status, msg, websock_state} ->
         do_deflate(msg, socket, %{connection | websock_state: websock_state})
 
@@ -239,14 +230,20 @@ defmodule Bandit.WebSocket.Connection do
         do_deflate(msg, socket, %{connection | websock_state: websock_state})
 
       {:stop, :normal, websock_state} ->
-        do_close(@normal_close_code, socket, %{connection | websock_state: websock_state})
+        do_close(@normal_close_code, <<>>, socket, %{connection | websock_state: websock_state})
+
+      {:stop, :normal, {close_code, close_msg}, websock_state} ->
+        do_close(close_code, close_msg, socket, %{connection | websock_state: websock_state})
+
+      {:stop, :normal, close_code, websock_state} ->
+        do_close(close_code, <<>>, socket, %{connection | websock_state: websock_state})
 
       {:stop, reason, websock_state} ->
         do_error(1011, reason, socket, %{connection | websock_state: websock_state})
     end
   end
 
-  defp do_close(code, msg \\ <<>>, socket, connection) do
+  defp do_close(code, msg, socket, connection) do
     if connection.state == :open do
       connection.websock.terminate(:normal, connection.websock_state)
       Socket.close(socket, code, msg)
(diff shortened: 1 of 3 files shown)


moogle19:
If these changes are okay with you @mtrudel , we should also update `websock` to reflect the newly added `:close` replies.

mtrudel:
I'm away from computers this week but will look at this in detail next week. Any chance this relates to the recent work I have pending on websock (no computer here, so genuinely asking)?

moogle19:
> I'm away from computers this week but will look at this in detail next week. Any chance this relates to the recent work I have pending on websock (no computer here, so genuinely asking)?

Yes, it is achieving the same result, but uses the `:reply` tuple instead of `:stop` (I chose that, because cowboy uses the same pattern), but I also though about using `:stop`.

I am still not quite sure which would be better.

mtrudel:
Ah yes, I see the difference now. I think the explicit `:stop` tuple support as just added to websock is probably the cleaner option - it makes the API semantics hew a bit closer to the underlying protocol semantics, specifically that sending a close frame is an explicit state change in the server, and affects what messages are (not) valid thereafter. Seems much more 'final' to do this via a `:stop` tuple to me. 

To figure out the underlying implementation in Cowboy for the `{:reply, {:close, .......` form was a bit of a spelunk, and really drove home to me how unclear of an API contract it is ('does it close the connection?' 'does it call any callbacks?', etc)? I'd prefer to sidestep that if at all possible.

mtrudel:
With your most recent changes doesn't this boil down to the same thing as #99 ?

moogle19:
> With your most recent changes doesn't this boil down to the same thing as #99 ?

Not sure how did not see your pull request 🙈
Yes, it is basically the same thing.

mtrudel:
LMAO sorry for the duplicate work! Care to take a quick gander at #99?

PR #107, opened 2023-03-09
Change close code guard to favour happy path

ryanwinchester:
Check two ranges instead of three.

### DIFF
--- a/lib/bandit/websocket/connection.ex
+++ b/lib/bandit/websocket/connection.ex
@@ -117,8 +117,8 @@ defmodule Bandit.WebSocket.Connection do
         # This is a bit of a subtle case, see RFC6455§7.4.1-2
         reply_code =
           case frame.code do
-            code when code in 0..999 or code in 1004..1006 or code in 1012..2999 -> 1002
-            _code -> 1000
+            code when code in 1000..1003 or code in 1007..1011 or code > 2999 -> 1000
+            _code -> 1002
           end
 
         do_stop(reply_code, :remote, socket, connection)


lib/bandit/websocket/connection.ex
```diff
@@ -117,8 +117,8 @@ defmodule Bandit.WebSocket.Connection do
         # This is a bit of a subtle case, see RFC6455§7.4.1-2
         reply_code =
           case frame.code do
-            code when code in 0..999 or code in 1004..1006 or code in 1012..2999 -> 1002
-            _code -> 1000
+            code when code in 1000..1003 or code in 1007..1011 -> 1000
```
mtrudel, line 120: We're seeing some autobahn failures on these cases in CI. I think it's because of different behaviour on codes above 2999. I think the last clause could just be `code >= 1012`?

ryanwinchester: Yeah, it's not clear how they should be handled either? Which makes it worse.

[7.4.2](https://www.rfc-editor.org/rfc/rfc6455#section-7.4.2).  Reserved Status Code Ranges

   0-999

      Status codes in the range 0-999 are not used.

   1000-2999

      Status codes in the range 1000-2999 are reserved for definition by
      this protocol, its future revisions, and extensions specified in a
      permanent and readily available public specification.

   3000-3999

      Status codes in the range 3000-3999 are reserved for use by
      libraries, frameworks, and applications.  These status codes are
      registered directly with IANA.  The interpretation of these codes
      is undefined by this protocol.

   4000-4999

      Status codes in the range 4000-4999 are reserved for private use
      and thus can't be registered.  Such codes can be used by prior
      agreements between WebSocket applications.  The interpretation of
      these codes is undefined by this protocol.

mtrudel:
Great work! Thanks for the PR!

ryanwinchester:
> Great work! Thanks for the PR!

Ha! It turned out to be not really matter much in the end, since there are still 3 guard clauses. At least one of them isn't a range, though...

Issue #101, opened 2023-02-23
`Plug.Conn.send_resp` in separate process fails for HTTP/2 streams


danschultzer:
Took me a while to track this one down. When testing with HTTP/2 I was seeing failure with `Stream 3 completed in unexpected state remote_closed` and empty responses even though I definitely did send the response!

The culprit is this line (the same for `send_data`):
https://github.com/mtrudel/bandit/blob/ebba77a60570de55b1150662c63513492fc14f6f/lib/bandit/http2/connection.ex#L365

In [TestServer](https://github.com/danschultzer/test_server) the route matching lives in a GenServer. When a requests hits the generic plug on the HTTP server, the plug will pass off the conn to the genserver to be processed in a `handle_call` callback. This means that caller pid will be different when calling `send_resp`.

Resolving this in TestServer may be awkward since the routing state lives in that GenServer process. Maybe it does make sense that the `send_resp` call should only happen in the generic plug process. But what's the reasoning behind the `owner?/2` check? Why is it necessary to lock the conn processing to the initial caller?

FWIW cowboy works with the current logic in TestServer. It works as expected with Bandit when I remove this check, but not sure if there are any sideeffects to doing that. If it's necessary to have this check then it would be good to have a clearer warning.

mtrudel:
That's an oddball one! I suppose the plug API doesn't specify anything explicitly about cross-process plug usage, but it conceptually seems like doing so could introduce a ton of gnarly concurrency issues.

danschultzer:
Great! Yeah, maybe this is just an extreme edge case, and it was never really meant to handle the conn outside of the owner process. I can't think of any other library that would do this other than bypass.

I looked at the plug sourcecode. The `:owner` attribute seems to only be used to pass messages of what has happened with the conn:

https://github.com/elixir-plug/plug/blob/d7940552e365e0d3245bb78ca99429ea5c324f7d/lib/plug/conn.ex#L447-L455
https://github.com/elixir-plug/plug/blob/d7940552e365e0d3245bb78ca99429ea5c324f7d/lib/plug/conn.ex#L486-L506
https://github.com/elixir-plug/plug/blob/d7940552e365e0d3245bb78ca99429ea5c324f7d/lib/plug/conn.ex#L528-L534

[...]

PR #115, opened 2023-03-25
Add warning for different owner access to stream

danschultzer:
Based on #101. This gives you better feedback for when the connection unexpectedly closes calling `send_resp` from another process. Without this you only see:

```
16:18:05.922 [warning] Header timestamp couldn't be fetched from ETS cache

16:18:05.925 [warning] Stream 3 completed in unexpected state remote_closed
```

That makes it difficult to understand what happens unless you go through the whole code path that includes GenServer messages (took me some time to figure out). With this PR you instead get these warnings:

```
16:18:05.922 [warning] Header timestamp couldn't be fetched from ETS cache

16:18:05.925 [warning] Stream 3 was handled by #PID<0.4048.0> but MUST be handled by #PID<0.4366.0>, the connection will complete unexpectedly

16:18:05.925 [warning] Stream 3 was handled by #PID<0.4048.0> but MUST be handled by #PID<0.4366.0>, the connection will complete unexpectedly

16:18:05.925 [warning] Stream 3 completed in unexpected state remote_closed
```

But the conn response is odd since status will be a nil value and empty body.

### DIFF
--- a/lib/bandit/http2/stream.ex
+++ b/lib/bandit/http2/stream.ex
@@ -219,7 +219,11 @@ defmodule Bandit.HTTP2.Stream do
 
   @spec owner?(t(), pid()) :: :ok | {:error, :not_owner}
   def owner?(%__MODULE__{pid: pid}, pid), do: :ok
-  def owner?(%__MODULE__{}, _pid), do: {:error, :not_owner}
+  def owner?(%__MODULE__{} = stream, pid) do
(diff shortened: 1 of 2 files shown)


mtrudel:
I like this code! But #117 would make this unnecessary.

danschultzer:
#117 looks good, closing this!

Issue #35, opened 2022-10-07
Multiple/dynamic SSL certificates?


carterbryden:
Hi, I'm considering building a pure elixir reverse proxy library and I'm looking at Bandit as a possible option to be the server at the core of it.

One of the main features would be on-demand SSL certificates for sites dynamically added through an API layer, using an acme protocol client (that I'll need to write).

Looking at the docs, SSL handling for Bandit currently seems to be locked to a single domain and cert. I'm wondering if you'd be open to the idea of allowing dynamic domains and certs that can be updated during runtime. If so, do you have any thoughts about how that should be handled and how best I could contribute to that?

Thanks!

moogle19:
Hi @carterbryden ,

The problem you describe is actually one (or two..) layers deeper, which means in the case of `bandit` in [thousand_island](https://github.com/mtrudel/thousand_island).

`thousand_island` is using the default Erlang/OTP `:ssl`.
`:ssl` allows you to define either a [sni_hosts](https://www.erlang.org/doc/man/ssl.html#type-sni_hosts) keyword list with a static `hostname => server_options` mapping or a [sni_fun](https://www.erlang.org/doc/man/ssl.html#type-sni_fun) which is called on every call containing a [SNI](https://en.wikipedia.org/wiki/Server_Name_Indication), which is probably what you want.

The `server_options` returned by the `sni_fun` are merged into the `server_options` defined in the `listen` function (overriding the cert / key defined there), which means you can set your dynamic `cert` / `certfile` + `key` / `keyfile` there.

You can define your custom `sni_fun` in the `transport_options` of bandit, like this:
```elixir
bandit_options = [
  port: 4000,
  transport_options: [
    sni_fun: &YourCustom.sni/1,
    certfile: "priv/certs/cert.pem",
    keyfile: "priv/certs/key.pem"
  ]
]

children = [
  {Bandit, plug: Some.Plug, scheme: :https, options: bandit_options}
]
```

mtrudel:
@moogle19 exactly! I've never done this as such, but this is exactly why I chose to not be clever about option parsing and instead just expose everything that OTP's `:gen_tcp` and `:ssl` applications provide (see https://github.com/mtrudel/thousand_island/issues/13 for prior discussion). Being able to stand on the shoulders of the OTP giants who implemented `sni_fun` should 'just work' for you here. 

Would love to hear about success stories with this!

mtrudel:
[...]

PR #116, opened 2023-03-26
Adjustable info log for server startup

danschultzer:
I want to be able to silence the info log that Bandit by default prints. The `:log` option is lifted from how it's usually dealt with in Plug and Phoenix.

But I wonder if there should be an info log by default? With `Plug.Cowboy` and `:httpd` there are no info log for server start. This is something that Phoenix adds to the Cowboy2 adapter. If you start the server yourself you would already be aware of the URL/port settings. It makes sense for higher level like Phoenix though.

So I think maybe this info log belongs in `Bandit.PhoenixAdapter` instead. It should maybe also have a pretty print of common errors just like the Cowboy2 adapter. Right now with `:eaddrinuse` it just shows this:

```
     ** (EXIT from #PID<0.13506.0>) shutdown: failed to start child: :listener
         ** (EXIT) :eaddrinuse
```

Phoenix instead pretty prints this (using the info private function):

```
[error] Running MyAppWeb.Endpoint with cowboy 2.9.0 at http failed, port already in use
```

### DIFF
--- a/lib/bandit.ex
+++ b/lib/bandit.ex
@@ -71,7 +71,7 @@ defmodule Bandit do
   * `scheme`: One of `:http` or `:https`. If `:https` is specified, you will need
      to specify `certfile` and `keyfile` in the `transport_options` subsection of `options`.
      Defaults to `:http`
-  * `:log` - Log level to use for Bandit start log event.
+  * `:startup_log` - The log level at which Bandit should log startup info.
     Defaults to `:info` log level, can be set to false to disable it.
   * `options`: Options to pass to `ThousandIsland`. For an exhaustive list of options see the
     `ThousandIsland` documentation, however some common options are:
@@ -216,7 +216,7 @@ defmodule Bandit do
     arg =
       arg
       |> validate_options(
-        ~w(scheme plug display_plug options http_1_options http_2_options websocket_options log)a,
+        ~w(scheme plug display_plug options http_1_options http_2_options websocket_options startup_log)a,
         "top level"
       )
 
@@ -254,7 +254,7 @@ defmodule Bandit do
 
     {plug_mod, _} = plug = plug(arg)
     display_plug = Keyword.get(arg, :display_plug, plug_mod)
-    log = Keyword.get(arg, :log, :info)
+    startup_log = Keyword.get(arg, :startup_log, :info)
 
     handler_options = %{
       plug: plug,
@@ -278,7 +278,7 @@ defmodule Bandit do
     |> ThousandIsland.start_link()
     |> case do
       {:ok, pid} ->
-        log && Logger.log(log, info(scheme, display_plug, pid))
+        startup_log && Logger.log(startup_log, info(scheme, display_plug, pid))
         {:ok, pid}
 
       {:error, _} = error ->
--- a/lib/bandit.ex
+++ b/lib/bandit.ex
@@ -281,6 +281,13 @@ defmodule Bandit do
         startup_log && Logger.log(startup_log, info(scheme, display_plug, pid))
         {:ok, pid}
 
+      {:error, {:shutdown, {:failed_to_start_child, :listener, :eaddrinuse}}} = error ->
+        port = Keyword.get(options, :port, 4000)
+        address = options |> Keyword.get(:transport_options, []) |> Keyword.get(:ip, {127, 0, 0, 1})
+
+        Logger.error([info(scheme, display_plug, %{port: port, address: address}), " failed, port already in use"])
+        error
+
       {:error, _} = error ->
         error
     end
@@ -306,14 +313,18 @@ defmodule Bandit do
     end
   end
 
-  defp info(scheme, plug, pid) do
+  defp info(scheme, plug, pid_or_info) do
     server_vsn = Application.spec(:bandit)[:vsn]
-    "Running #{inspect(plug)} with Bandit #{server_vsn} at #{bound_address(scheme, pid)}"
+    "Running #{inspect(plug)} with Bandit #{server_vsn} at #{bound_address(scheme, pid_or_info)}"
   end
 
-  defp bound_address(scheme, pid) do
-    {:ok, %{address: address, port: port}} = ThousandIsland.listener_info(pid)
+  defp bound_address(scheme, pid) when is_pid(pid) do
+    {:ok, info} = ThousandIsland.listener_info(pid)
 
+    bound_address(scheme, info)
+  end
+
+  defp bound_address(scheme, %{address: address, port: port}) do
     case address do
       {:local, unix_path} ->
         "#{unix_path} (#{scheme}+unix)"
--- a/lib/bandit.ex
+++ b/lib/bandit.ex
@@ -282,10 +282,7 @@ defmodule Bandit do
         {:ok, pid}
 
       {:error, {:shutdown, {:failed_to_start_child, :listener, :eaddrinuse}}} = error ->
-        port = Keyword.get(options, :port, 4000)
-        address = options |> Keyword.get(:transport_options, []) |> Keyword.get(:ip, {127, 0, 0, 1})
-
-        Logger.error([info(scheme, display_plug, %{port: port, address: address}), " failed, port already in use"])
+        Logger.error("Startup failed; address/port already in use")
         error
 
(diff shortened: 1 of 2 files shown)


mtrudel:
(Ugh, I had a whole multiple paragraph bit of response here and Safari ate it. The gist of it was that I feel pretty strongly that it's Bandit's job to be logging things as it does today, and that I don't think that moving the responsibility into the Phoenix Adapter furthers Bandit's fundamental goal of clarity).

That being said, I like this change a lot and am happy to see it land! A couple of things though; 

* I wonder if we should call it `log_startup` to reflect the fact that it only affects startup logs. As it's named now, folks might think it would affect logging everywhere within the app
* I'd be happy to see an extra clause added to the case block around `bandit.ex:289` to handle address in use errors as you suggest!

lib/bandit.ex
```diff
@@ -275,9 +278,16 @@ defmodule Bandit do
     |> ThousandIsland.start_link()
     |> case do
       {:ok, pid} ->
-        Logger.info(info(scheme, display_plug, pid))
+        startup_log && Logger.log(startup_log, info(scheme, display_plug, pid))
         {:ok, pid}
 
+      {:error, {:shutdown, {:failed_to_start_child, :listener, :eaddrinuse}}} = error ->
+        port = Keyword.get(options, :port, 4000)
+        address = options |> Keyword.get(:transport_options, []) |> Keyword.get(:ip, {127, 0, 0, 1})
```
mtrudel, line 286: This is expressing policy that's part of Thousand Island (specifically, default ports & bind addresses). We should really be pulling this out of config somehow - let me tinker with this and see what I can come up with

danschultzer: Yup, didn't find a better way to pull this info, so just guessing here from the opts. The Cowboy2 adapter in Phoenix uses the ranch ref to get info from a local ETS table. I don't know if there's any way to get the info from Bandit once the supervision tree has been terminated? Feel free to continue work on this PR!

lib/bandit.ex
```diff
@@ -71,6 +71,8 @@ defmodule Bandit do
   * `scheme`: One of `:http` or `:https`. If `:https` is specified, you will need
      to specify `certfile` and `keyfile` in the `transport_options` subsection of `options`.
      Defaults to `:http`
+  * `:startup_log` - The log level at which Bandit should log startup info.
```
danschultzer, line 74: I've updated naming to `startup_log`. It seemed better than `log_startup` since it's the log level, not boolean.

danschultzer:
> I don't think that moving the responsibility into the Phoenix Adapter furthers Bandit's fundamental goal of clarity

Yeah, as long as it's possible to silence the info log then it's all good. But I want to expand a little on my thoughts here.

When I manually start Bandit I'm aware of the status and opts since it's all right there. I don't think I need a log entry to know that the server has started or what the address/port is, it's already given?

However, the info log becomes very necessary when there's any abstraction, i.e. like it being part of the Phoenix supervision tree. In that case I wouldn't necesarily know anything about the Bandit server or how to access it.

Here's an example from the [Plug readme](https://hexdocs.pm/plug/1.14.2/readme.html#hello-world-request-response) with the `Plug.Cowboy` server:

```elixir
require Logger
webserver = {Plug.Cowboy, plug: Router, scheme: :http, port: 4000}
{:ok, _} = Supervisor.start_link([webserver], strategy: :one_for_one)
Logger.info("Plug now running on localhost:4000")
```

If we imagine that `Plug.Cowboy` was spitting out the info log in the child process inside the supervisor supervision tree, would that really make sense rather than handled outside the supervision tree? You could imagine there's a bunch of children there that all must start/warm up together before printing a success message.

Obviously not a big deal, you can just deal with this by reordering the supervision tree so all the required children will fail before it gets to Bandit. Another counter point is that default/implicit port and address makes it more necessary to info log it at low level.

Again it's all good with ability to silence the info log, just some thoughts on why it should be there in the first place 😄

lib/bandit.ex
mtrudel: > Again it's all good with ability to silence the info log, just some thoughts on why it should be there in the first place

Yep, totally agree - I think the config point we're adding here handles this well. I really *do* like having the logging by default for all Bandit installs; I personally use it all the time just to double check myself.

mtrudel:
> Again it's all good with ability to silence the info log, just some thoughts on why it should be there in the first place

Yep, totally agree - I think the config point we're adding here handles this well. I really *do* like having the logging by default for all Bandit installs; I personally use it all the time just to double check myself.

mtrudel:
Ended up significantly simplifying the error message in this case. I've been [on the record](https://github.com/mtrudel/thousand_island/issues/13#issuecomment-1032940131) before (also [here](https://github.com/mtrudel/bandit/issues/35#issuecomment-1271674797) about the tradeoffs in in (not) wrapping the underlying libraries' error responses and overall concepts, and this is another one where I think that 'not trying to be too clever' is the right thing to do. 

There are just too many corner cases with trying to divine what exact address / port the user is binding to for me to feel comfortable with us trying to tell the user anything too specific. Even within Thousand Island, we don't necessarily know the bind address / port that would end up being used (for example, if the user specifies port 0, it's up the the OS to find a port for us. The default bind address may change between TCP and SSL, and so on). Rather than getting on a treadmill of trying to cover off all the possible cases, I think it's best to be transparent and unobtrusive with the underlying error. 

WDYT?

danschultzer:
Yeah that makes sense. LGTM! 🚢 

Only alternative I see here would be to ensure the port/address is available in the caller process somehow (like how ranch does it with ETS), but would be way overkill and a rabbit hole to just have the ability to print address/port in the error message.

mtrudel:
> Only thing I could think of here would be to ensure the port/address is available in the caller process somehow (like how ranch does it with ETS), but would be way overkill and a rabbit hole to just have the ability to print address/port in the error message. And you will end up in the same situation of trying to deal with edge cases.

That's the thing though - the eaddrinuse error comes from [here](https://github.com/mtrudel/thousand_island/blob/main/lib/thousand_island/listener.ex#L30), and in that case, we simply don't know what the actual resolved bind address/port was that failed (we do know what we passed into the transport module as options, but it's entirely possible that the transport module overrode those / set defaults that aren't really spelled out in authoritative detail anywhere in the Erlang docs). You can really only know the bind info after a successful bind; it's fundamentally a chicken and egg scenario. 

Reviewing ranch's code for this [here](https://github.com/ninenines/ranch/blob/25074597ef7c8cbf36319a3170fd3e787516dc12/src/ranch_acceptors_sup.erl#L68) and elsewhere further down in that file, it's not clear to be that ranch is ever in a situation to meaningfully provide bind data for a ref corresponding to a server that failed to start (it reads as if it [exits gracelessly](https://github.com/ninenines/ranch/blob/25074597ef7c8cbf36319a3170fd3e787516dc12/src/ranch_acceptors_sup.erl#L96) and indeed this is what seems to happen if you try and start two out of the box phoenix apps on the same port. 

I wonder if perhaps the error lines in CowboyAdapter never actually match? I can't seem to get them to.....

danschultzer:
You are right. Didn't realize that the address and port is never shown (I even have this as an example in the issue). It just catches the arg error and [returns the scheme](https://github.com/phoenixframework/phoenix/blob/b299b8639d1c730ae2734275bcd78d5bc963d8ce/lib/phoenix/endpoint/cowboy2_adapter.ex#L124-L134):

```elixir
  defp bound_address(scheme, ref) do
    case :ranch.get_addr(ref) do
      {:local, unix_path} ->
        "#{unix_path} (#{scheme}+unix)"

      {addr, port} ->
        "#{:inet.ntoa(addr)}:#{port} (#{scheme})"
    end
  rescue
    _ -> scheme
  end
```

So nevermind, pretty sure it'll never show up if the address/port is not used, and is only ever used for when it succeeds.

danschultzer:
Pushed a small change so it conforms the message by handling nil pid instead. I should have looked at it more carefully, it should probably be updated in Cowboy2. Maybe there's a possibility that the address/port fails to fetch even when it's running and that's why it's handled in such an odd way.

danschultzer:
@mtrudel just amended change to fix linter error.

mtrudel:
Thanks for the PR!

PR #122, opened 2023-03-28
Require Elixir 1.12

danschultzer:
Got an error testing with Elixir 1.11, because of `Kernel.then/2` usage.

`mtrudel/elixir-ci-actions/.github/workflows/test.yml@main` tests from 1.12 so this was not caught, maybe it should explicitly test the lowest version supported just to ensure that there's no drift in the CI?

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -5,7 +5,7 @@ defmodule Bandit.MixProject do
     [
       app: :bandit,
       version: "0.7.4",
-      elixir: "~> 1.11",
+      elixir: "~> 1.12",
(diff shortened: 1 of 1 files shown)


danschultzer:
Hmm, seems like there's a flakey test, probably means that the startup message should be waited on, or should be silenced?

mtrudel:
> Hmm, seems like there's a flakey test, probably means that the startup message should be waited on, or should be silenced?

Yep. Just fixed in main.

mtrudel:
Thanks for the PR!

PR #123, opened 2023-03-28
Fix test for OTP 26

moogle19:
OTP 26 made some changes to maps, so they aren't (predictably) sorted anymore (see [here](https://www.erlang.org/news/162#maps)).

This fixes a failing test for that case.


mtrudel:
Nice! I'm going to get OTP 26 into the test matrix today as well.

There's a [machete matcher](https://hexdocs.pm/machete/Machete.InAnyOrderMatcher.html) specifically for this, so I just rewrote the test to use it instead of relying on a manual sort

mtrudel:
@moogle19 can you confirm this works for you on 26.0?

moogle19:
@mtrudel Yes, it works!

PR #128, opened 2023-04-11
Remove `:otp_app` after SSL configuration

moogle19:
`Plug.SSL.configure/1` [requires the `:otp_app` key,](https://github.com/elixir-plug/plug/blob/2ef07cdd2732cde5cac73fc39b49fe83d5fcc369/lib/plug/ssl.ex#L226-L232) when using relative certificate paths.

If the `:otp_app` key is set, it is kept in the `transport_options` of the `ThousandIsland.ServerConfig`, which results in a `:badarg` when `:ssl.listen/2` is called [here](https://github.com/mtrudel/thousand_island/blob/ed8dff6d5c9bf147bbe06e3e8e501b6e0654fc12/lib/thousand_island/listener.ex#L12).

`Plug.Cowboy` does the same thing [here](https://github.com/elixir-plug/plug_cowboy/blob/1796e77aa42c2284f3667c7aeac7074236f47c07/lib/plug/cowboy.ex#L355).

I am not sure why this field is kept in the config and not removed by `Plug.SSL.configure/1`.

### DIFF
--- a/lib/bandit.ex
+++ b/lib/bandit.ex
@@ -109,7 +109,7 @@ defmodule Bandit do
     valid `certfile` and `keyfile` values (or an equivalent value within
     `thousand_island_options.transport_options`). Defaults to `:http`
   * `port`: The TCP port to listen on. This option is offered as a convenience and actually sets
-    the option of the same name within `thousand_island_options`. If ia string value is passed, it 
+    the option of the same name within `thousand_island_options`. If ia string value is passed, it
     will be parsed as an integer. Defaults to 4000 if `scheme` is `:http`, and 4040 if `scheme` is
     `:https`
   * `ip`:  The interface(s) to listen on. This option is offered as a convenience and actually sets the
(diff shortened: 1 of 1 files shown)


mtrudel:
I'm sure I'd tested this. Do you have a working demo?

mtrudel:
More colour: I'd actually originally wanted to put the Plug.SSL.configure call into thousand island but abandoned that idea when I realized thousand island would have to have a dependency on Plug. If you ask me, I think that Plug is the wrong home for this code in the first place, and having it in Bandit is a concession. Oh well.

moogle19:
Here is a demo project for the problem: https://github.com/moogle19/bandit_ssl_test

It fails with:
```elixir
** (Mix) Could not start application bandit_ssl_test: BanditSslTest.Application.start(:normal, []) returned an error: shutdown: failed to start child: {Bandit, #Reference<0.827849923.2524971016.133297>}
    ** (EXIT) shutdown: failed to start child: :listener
        ** (EXIT) :badarg
```

mtrudel:
Excellent call on both points @moogle19 . Fixes were trivial and pushed directly to main at ba1f491738fb23b875271d46259a642f6eff4a33 and e86c4b2b1780dc9de08d7ea95b883740ecc8d961

Thanks for the attention to detail!

mtrudel:
Ah shoot! I didn't even notice that this was a PR; I thought it was an issue and didn't see your fix attached!

PR #136, opened 2023-04-26
Use valid path to avoid false negative

moogle19:
When testing for invalid version, we should use a valid path. 
Otherwise the assert wouldn't fail when the version check fails, since the invalid path also returns `400` (with "Unsupported request target (RFC9112§3.2)")


mtrudel:
Good catch! This was likely copy-pasting gone wrong.

PR #137, opened 2023-04-26
Swap out Finch for Req in tests

mtrudel:
Use Req instead of Finch as our client of choice. More concise & better API for our use case.

One open problem remains: we get error messages like the following when running `test/bandit/http2/plug_test.exs`:

```
15:20:21.066 [error] Failed to connect to https://localhost:53809: connection refused
```

(You can see this for yourself by checking out this branch and running `mix test test/bandit/http2/plug_test.exs`).

These errors only seem to occur on HTTP2 connections, and don't have any effect on the underlying test passing. I've managed to narrow this down to the fact that we use a single shared Finch pool; this seems to cause connections to error out as above. I'm not sure if this is because the pool keeps connections open after the test is complete and then gets surprised when the server shuts down, or if it's something else.

For illustration, I can fix this by moving to a per-process finch pool like so (apologies for the hackiness of this approach; it's simply a proof of concept):

```elixir
defmodule ReqHelpers do
  @moduledoc false

  defmacro __using__(_) do
    quote location: :keep do
      def req_http1_client(context) do
        [req: context |> req_opts() |> Req.new()]
      end

      def req_h2_client(context) do
        finch_name = self() |> inspect() |> String.to_atom()
        opts = [ name: finch_name, pools: %{ default: [ size: 50, count: 1, 
            protocol: :http2,
            conn_opts: [
            transport_opts: [
              verify: :verify_peer,
              cacertfile: Path.join(__DIR__, "../support/ca.pem")
            ]]
          ]
        ]
        {:ok, _} = start_supervised({Finch, opts})

        [req: context |> req_opts() |> Keyword.put(:finch, finch_name) |> Req.new()]
      end

      defp req_opts(context) do
        [
          base_url: context.base,
          retry: false,
          compressed: false,
          raw: true,
          follow_redirects: false
        ]
      end
    end
  end
end
```

@wojtekmach do you have any ideas or suggestions here?


mtrudel:
> yeah, that's exactly it. It took me some grepping for "Failed to connect to" but yeah it was indeed Finch.HTTP2.Pool that logged these.

Ah yeah, sorry. I'd noticed that as well but neglected to mention it.

> If there's an option to disable re-connecting then that'd probably be the easiest way to fix it. Otherwise, I think I'd start a finch pool per test.

It doesn't seem that there's an option for that in Finch - I'll likely take the per-test pool approach for now. 

> Btw, always great to see Req used in the wild! ❤️

It's great to use! Really amazing work!

> I noticed you're disabling a bunch of options which totally makes sense for your use case. What could potentially make sense too is disabling all built-in steps which would mean Req would be just tiny layer above Finch. You can do that by switching `Req.new() |>`  to `%Req.Request{} |>`

I will try that! In my case I want the requests to be as ''dumb' as possible (ie: all of the Req magic for handling compression, etc actually works against me). This seems to make more sense.

PR #138, opened 2023-04-28
Minor performance improvement

moogle19:
I think we should avoid a nested lookup on every connection and set the `enabled` flags directly in the `handler_options` / `state`. 

WDYT?

### DIFF
--- a/lib/bandit.ex
+++ b/lib/bandit.ex
@@ -109,7 +109,7 @@ defmodule Bandit do
     valid `certfile` and `keyfile` values (or an equivalent value within
     `thousand_island_options.transport_options`). Defaults to `:http`
   * `port`: The TCP port to listen on. This option is offered as a convenience and actually sets
-    the option of the same name within `thousand_island_options`. If ia string value is passed, it 
+    the option of the same name within `thousand_island_options`. If ia string value is passed, it
     will be parsed as an integer. Defaults to 4000 if `scheme` is `:http`, and 4040 if `scheme` is
     `:https`
   * `ip`:  The interface(s) to listen on. This option is offered as a convenience and actually sets the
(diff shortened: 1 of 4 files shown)


lib/bandit.ex
```diff
@@ -321,7 +321,9 @@ defmodule Bandit do
     handler_options = %{
       plug: plug,
       handler_module: Bandit.InitialHandler,
-      opts: %{http_1: http_1_options, http_2: http_2_options, websocket: websocket_options}
+      opts: %{http_1: http_1_options, http_2: http_2_options, websocket: websocket_options},
+      http_1_enabled: Keyword.get(http_1_options, :enabled, true),
+      http_2_enabled: Keyword.get(http_2_options, :enabled, true)
     }
```
ryanwinchester, line 327: Do they need to stay in the original set of options then?

```suggestion
{http_1_enabled, http_1_options} = Keyword.pop(http_1_options, :enabled, true)
{http_2_enabled, http_2_options} = Keyword.pop(http_2_options, :enabled, true)

handler_options = %{
  plug: plug,
  handler_module: Bandit.InitialHandler,
  opts: %{http_1: http_1_options, http_2: http_2_options, websocket: websocket_options},
  http_1_enabled: http_1_enabled,
  http_2_enabled: http_2_enabled
}
```

moogle19: I am not quite sure. The `opts` get passed down to the `Adapter` but it seem the `enabled` field is only used for websockets during upgrade attempts and not for http1 or http2, so we could safely remove them from the `opts`.

mtrudel (approved):

mtrudel:
@moogle19 lmk if you want to delete the values or not, and I'll get this merged either way!

moogle19:
If you don't see a problem let's delete them.

moogle19:
I updated the PR with the suggestions from @ryanwinchester and added the same pattern for the websocket options.

mtrudel:
Looks great!

PR #142, opened 2023-05-12
Update README.md

cvkmohan:
A copy paste of the line is causing an error. So, the change is for lazy people like me. :)


mtrudel:
Good catch! The right option here though is `~> 1.0-pre`, not `>= 1.0.0-pre`. Updated as such and will be merged directly!

Thanks for the PR!

PR #143, opened 2023-05-15
Use `byte_size/1` instead of `String.length/1`

moogle19:
The documentation states:

> max_header_key_length: The maximum permitted length of any single header key (expressed as the number of decompressed bytes) in an HTTP/2 request. Defaults to 10_000 bytes

This isn't the case with `String.length/1` (which is also quite a bit slower than `byte_size/1`) since it counts multi-byte characters as 1.


mtrudel:
Looks like a decent perf lift from this! Great catch!

PR #146, opened 2023-05-17
Add credo config file

ryanwinchester:
Make it easier to configure how `credo` is run.

Example: CI uses `--strict`, so in the config I turned `strict` on by default to avoid running `mix credo` locally and thinking all is well before pushing.

Also added custom config for `Credo.Check.Design.AliasUsage` so we could remove some magic credo comments.

i.e.

```elixir
[
  if_called_more_often_than: 1,
  excluded_lastnames: defaults ++ ~w[Adapter Utils]
]
```

or could just disable the alias check altogether?

Could potentially do similar for `Credo.Check.Refactor.CyclomaticComplexity`.

### DIFF
--- a/.credo.exs
+++ b/.credo.exs
@@ -1,0 +1,216 @@
+# This file contains the configuration for Credo and you are probably reading
+# this after creating it with `mix credo.gen.config`.
+#
+# If you find anything wrong or unclear in this file, please report an
+# issue on GitHub: https://github.com/rrrene/credo/issues
+#
+%{
+  #
+  # You can have as many configs as you like in the `configs:` field.
+  configs: [
+    %{
+      #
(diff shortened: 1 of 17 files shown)


mtrudel:
Yeah! This is great! I've always hated how credo exclusions just kinda end up everywhere. Consolidating it makes a lot of sense.

I got rid of all the in-file credo overrides:

* Cyclomatic complexity now checks globally at 15 (it's a sensible check, just a bit too sensitive)
* I disabled alias checks entirely; I have a soft policy of not using aliases in public code to make it more explicit and require less ramp-up speed to read, so disabling this check wholesale makes sense in this project

Have a look and if you're happy with these changes we can merge this straightaway!

PR #145, opened 2023-05-17
Port number guard checks valid port range

ryanwinchester:
Should it check that `port` matches `t:inet.port_number/0`?

PR #147, opened 2023-05-17
Add a port number guard and use that in the checks instead of range

ryanwinchester:
I talked about it at the end of #145 although it's a micro-optimization.

I didn't know where to put it, I could put it as a private guard in the headers module? Or somewhere else?

### DIFF
--- a/lib/bandit/headers.ex
+++ b/lib/bandit/headers.ex
@@ -2,6 +2,8 @@ defmodule Bandit.Headers do
   @moduledoc false
   # Conveniences for dealing with headers
 
+  defguardp is_port_number(port) when Bitwise.band(port, 0xFFFF) === port
+
   def get_header(headers, header) do
     case List.keyfind(headers, header, 0) do
       {_, value} -> value
@@ -16,7 +18,7 @@ defmodule Bandit.Headers do
     |> case do
       [host, port] ->
         case Integer.parse(port) do
-          {port, ""} when port in 0..65_535 -> {:ok, host <> "]", port}
+          {port, ""} when is_port_number(port) -> {:ok, host <> "]", port}
           _ -> {:error, "Header contains invalid port"}
         end
 
@@ -31,7 +33,7 @@ defmodule Bandit.Headers do
     |> case do
       [host, port] ->
         case Integer.parse(port) do
-          {port, ""} when port in 0..65_535 -> {:ok, host, port}
+          {port, ""} when is_port_number(port) -> {:ok, host, port}
           _ -> {:error, "Header contains invalid port"}
         end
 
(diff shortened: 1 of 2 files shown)


mtrudel:
This is an extremely hot path (one of the hottest), so though it's a micro optimization it's an important one!

lib/bandit/ports.ex
```diff
@@ -0,0 +1,21 @@
+defmodule Bandit.Ports do
+  @moduledoc """
+  Bandit helper functions and guards for dealing with port numbers.
+  """
```
mtrudel, line 4: Internal libs like this use `@moduledoc false` to keep the docs focused on user-visible parts of Bandit.

lib/bandit/ports.ex
```diff
@@ -0,0 +1,21 @@
...
+
+  ## Examples
+
+      iex> is_port_number(443)
+      true
+
+      iex> is_port_number(-1)
+      false
+
+  """
```
mtrudel, line 19: ibid.

lib/bandit/headers.ex
```diff
@@ -2,6 +2,8 @@ defmodule Bandit.Headers do
   @moduledoc false
   # Conveniences for dealing with headers
 
+  import Bandit.Ports
+
```
mtrudel, line 6: The implementation of this is pretty simple; I think this is a rare case of `defguardp`

test/bandit/http1/request_test.exs
```diff
@@ -108,6 +108,64 @@ defmodule HTTP1RequestTest do
       assert Jason.decode!(body)["port"] == 1234
     end
 
+    test "derives port from host header for all valid ports", context do
+      client = SimpleHTTP1Client.tcp_client(context)
+
+      for port <- 0..65_535 do
```
mtrudel, line 114: This is a lot of requests! We could probably stand to do a subset of this (either random, some set of pre-selected ones, or using a `//step` on the range

ryanwinchester: I kind of thought that's what you intended when you wanted my other tests ported to these modules, but I pushed a more sane version.

lib/bandit/headers.ex
```diff
@@ -2,6 +2,8 @@ defmodule Bandit.Headers do
   @moduledoc false
   # Conveniences for dealing with headers
 
+  defguardp is_port_number(port) when Bitwise.band(port, 0xFFFF) === port
+
```
mtrudel, line 6: Nit, but if this is private oughtn't it be at the bottom of the file?

ryanwinchester: guards have to be defined above where you use them

ryanwinchester: ^

ryanwinchester:
I've been playing around with another micro-optimization for this module. Another PR and benchmarks incoming when this is merged.

PR #155, opened 2023-06-02
Add dark mode logo version in README

nelsonmestevao:
Learn more about this GitHub feature [here](https://github.blog/changelog/2021-11-24-specify-theme-context-for-images-in-markdown/).


mtrudel:
Hrm. This is showing up with broken images in my (light mode) Safari: 

image

nelsonmestevao:
That it's because I am pointing to images that would be in the main of this repository, and the image with light text is not there yet.

But it will work like the [Elixir repo does](https://github.com/elixir-lang/elixir). Example in darkmode (the dark text does not show):

image

mtrudel:
Ah, yes that makes sense!

nelsonmestevao:
I am sorry, I will try to figure out what's wrong :pray:

ryanwinchester:
So much better, thanks

PR #156, opened 2023-06-02
Fix README image duplication

nelsonmestevao:
The GitHub flavored Markdown does not work well with links. I think the links are nice but not necessary in this case, so I removed them in order to be able to have light and dark mode logos.


nelsonmestevao:
Preview available [here](https://github.com/nelsonmestevao/bandit/blob/ne/logo-light/README.md).

mtrudel:
Works for me! Thanks again!

PR #158, opened 2023-06-07
WIP: Extract Adapter.read/5 routine into tight read loop and add typespecs

asakura:
It seems that there is a way to speed up the read/5 function by reducing number of reduction it produces during execution.

```
Name                ips        average  deviation         median         99th %
read            17.65 K       56.64 μs    ±19.15%       53.83 μs       75.33 μs
read_with        5.86 K      170.73 μs    ±26.58%      163.62 μs      274.65 μs

Name      Reduction count
read               3.51 K
read_with         12.50 K - 3.56x reduction count +8.99 K
```

The benchmark in form of a livebook, assembly code and the rest of results you can find here: https://gist.github.com/asakura/8aabd3fed8859b36949f2df9d02e685a

### DIFF
--- a/lib/bandit/http1/adapter.ex
+++ b/lib/bandit/http1/adapter.ex
@@ -286,22 +286,46 @@ defmodule Bandit.HTTP1.Adapter do
   # Internal Reading
   ##################
 
-  @dialyzer {:no_improper_lists, read: 5}
-  defp read(socket, to_read, opts \\ [], bytes_read \\ 0, already_read \\ []) do
+  @spec read(
+          ThousandIsland.Socket.t(),
+          non_neg_integer(),
+          [read_length: non_neg_integer(), read_timeout: timeout()],
+          non_neg_integer(),
+          iodata()
+        ) ::
+          {:ok, iolist()}
+          | {:error, :closed | :timeout | :inet.posix()}
+  defp read(socket, to_read, opts \\ [], _bytes_read \\ 0, already_read \\ []) do
     read_size = min(to_read, Keyword.get(opts, :read_length, 1_000_000))
     read_timeout = Keyword.get(opts, :read_timeout)
+    do_read(socket, read_size, read_timeout, to_read, already_read)
+  end
 
-    with {:ok, chunk} <- ThousandIsland.Socket.recv(socket, read_size, read_timeout) do
-      remaining_bytes = to_read - byte_size(chunk)
-      bytes_read = bytes_read + byte_size(chunk)
+  @dialyzer {:no_improper_lists, do_read: 5}
+  @spec do_read(
+          ThousandIsland.Socket.t(),
+          non_neg_integer(),
+          timeout(),
+          non_neg_integer(),
+          iolist()
+        ) ::
+          {:ok, iolist()}
+          | {:error, :closed | :timeout | :inet.posix()}
+  defp do_read(socket, read_size, read_timeout, remaining_bytes, already_read)
+       when remaining_bytes > 0 do
+    case ThousandIsland.Socket.recv(socket, read_size, read_timeout) do
+      {:ok, chunk} ->
+        do_read(socket, read_size, read_timeout, remaining_bytes - byte_size(chunk), [
+          already_read | chunk
+        ])
 
-      if remaining_bytes > 0 do
-        read(socket, remaining_bytes, opts, bytes_read, [already_read | chunk])
-      else
-        {:ok, [already_read | chunk]}
-      end
+      err ->
+        err
     end
   end
+
+  defp do_read(_socket, _read_size, _read_timeout, _remaining_bytes, already_read),
+    do: {:ok, already_read}
 
   ##################
   # Response Sending
--- a/lib/bandit/http1/adapter.ex
+++ b/lib/bandit/http1/adapter.ex
@@ -298,7 +298,10 @@ defmodule Bandit.HTTP1.Adapter do
   defp read(socket, to_read, opts \\ [], _bytes_read \\ 0, already_read \\ []) do
     read_size = min(to_read, Keyword.get(opts, :read_length, 1_000_000))
     read_timeout = Keyword.get(opts, :read_timeout)
-    do_read(socket, read_size, read_timeout, to_read, already_read)
+    # when to_read is 0, we still want to read at least 1 byte
+    # or everything which is already in the buffer of the socket
+    remaining_bytes = max(to_read, 1)
+    do_read(socket, read_size, read_timeout, remaining_bytes, already_read)
   end
 
   @dialyzer {:no_improper_lists, do_read: 5}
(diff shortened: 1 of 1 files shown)


asakura:
BTW the lint job fails with funny reason:
```
:dialyzer.run error: Old PLT file /home/runner/work/bandit/bandit/priv/plts/dialyzer.plt
```

moogle19:
It seems the performance improvement is achieved by avoiding to access the `opts` in every function call.

If we rewrite the existing `read` function to something like this, we achieve equal performance numbers (and 1 reduction less 😄):

```elixir
defmodule ReadWith do
  def read(socket, to_read, opts \\ [], already_read \\ []) do
    read_size = min(to_read, Keyword.get(opts, :read_length, 1_000_000))
    read_timeout = Keyword.get(opts, :read_timeout)
    remaining_bytes = max(to_read, 1)

    do_read(socket, remaining_bytes, already_read, read_size, read_timeout)
  end

  defp do_read(socket, to_read, already_read, read_size, read_timeout) do
    with {:ok, chunk} <- ThousandIsland.Socket.recv(socket, read_size, read_timeout) do
      remaining_bytes = to_read - byte_size(chunk)

      if remaining_bytes > 0 do
        do_read(socket, remaining_bytes, [already_read | chunk], read_size, read_timeout)
      else
        {:ok, [already_read | chunk]}
      end
    end
  end
end
```

<details>
  <summary>Benchmark Results</summary>

  ```
##### With input Bigger #####
Name                ips        average  deviation         median         99th %
read_with      229.71 K        4.35 μs   ±122.37%        4.21 μs        7.71 μs
read           213.29 K        4.69 μs   ±119.91%        4.54 μs        8.08 μs

Comparison:
read_with      229.71 K
read           213.29 K - 1.08x slower +0.34 μs

Memory usage statistics:

Name         Memory usage
read_with        15.65 KB
read             15.65 KB - 1.00x memory usage +0 KB

Reduction count statistics:

Name      Reduction count
read_with          3.51 K
read               3.51 K - 1.00x reduction count +0.00100 K

**All measurements for reduction count were the same**

##### With input Micro #####
Name                ips        average  deviation         median         99th %
read             9.75 M      102.54 ns ±12767.57%          83 ns         208 ns
read_with        9.69 M      103.21 ns ±19875.14%          83 ns         167 ns

Comparison:
read             9.75 M
read_with        9.69 M - 1.01x slower +0.67 ns

Memory usage statistics:

Name         Memory usage
read                184 B
read_with           184 B - 1.00x memory usage +0 B

Reduction count statistics:

Name      Reduction count
read                   41
read_with              40 - 0.98x reduction count -1

**All measurements for reduction count were the same**

##### With input Small #####
Name                ips        average  deviation         median         99th %
read_with        1.99 M      502.36 ns  ±3108.22%         458 ns         584 ns
read             1.87 M      535.10 ns  ±3005.37%         500 ns         625 ns

Comparison:
read_with        1.99 M
read             1.87 M - 1.07x slower +32.74 ns

Memory usage statistics:

Name         Memory usage
read_with         1.59 KB
read              1.59 KB - 1.00x memory usage +0 KB

Reduction count statistics:

Name      Reduction count
read_with             329
read                  330 - 1.00x reduction count +1

**All measurements for reduction count were the same**

  ```
</details>

asakura:
Benchmark's results acquired through testing in the same controllable environment:
```
Name                 ips        average  deviation         median         99th %
read_with2       19.56 K       51.12 μs    ±64.54%       48.22 μs       70.01 μs
read             14.12 K       70.80 μs    ±87.12%       67.52 μs       97.27 μs
read_with         5.87 K      170.42 μs    ±14.50%      163.03 μs      277.13 μs

Comparison: 
read_with2       19.56 K
read             14.12 K - 1.39x slower +19.68 μs
read_with         5.87 K - 3.33x slower +119.30 μs

Name       Reduction count
read_with2          3.51 K
read                5.62 K - 1.60x reduction count +2.11 K
read_with          12.50 K - 3.56x reduction count +8.99 K
```

So yes, not calling `min/2` BIF allows to bump number of IPS and reduce reductions count. Regarding using `with` expressions: it compiles to absolutely same byte code (at least I couldn't spot difference) so there is no magic behind it.

asakura:
It seems that moving already_read acc left or right in argument list of `defp do_read(socket, to_read, already_read, read_size, read_timeout)` drops performance up to 16%. Does any one knows what causing this?

mtrudel:
PLT error is fixed - on occasion the Github CI setup wedges on a stale PLT cache & I have to clear it manually.

lib/bandit/http1/adapter.ex
```diff
@@ -286,23 +286,55 @@ defmodule Bandit.HTTP1.Adapter do
...
+  @spec read(
+          ThousandIsland.Socket.t(),
+          non_neg_integer(),
+          [read_length: non_neg_integer(), read_timeout: timeout()],
+          non_neg_integer(),
+          iodata()
+        ) ::
+          {:ok, iolist()}
+          | {:error, :closed | :timeout | :inet.posix()}
+  defp read(socket, to_read, opts \\ [], _bytes_read \\ 0, already_read \\ []) do
```
mtrudel, line 298: We can just get rid of bytes_read entirely I think?

lib/bandit/http1/adapter.ex
```diff
@@ -286,23 +286,55 @@ defmodule Bandit.HTTP1.Adapter do
...
     read_size = min(to_read, Keyword.get(opts, :read_length, 1_000_000))
     read_timeout = Keyword.get(opts, :read_timeout)
+    # when to_read is 0, we still want to read at least 1 byte
+    # or everything which is already in the buffer of the socket
+    remaining_bytes = max(to_read, 1)
```
mtrudel, line 303: Are you sure this won't end up with an off by one error in the case where we happen to get a read chunk that completely satisfies our desired length?

asakura: Please ignore it. `defmodule ReadWith3 do` contains the latest and greats implementation actually.

asakura:
Unfortunately the version suggested by @moogle19 won't work: it would try to read more than client has to transmit which leads to different broken tests due to `:timeout`. So `min/2` has to stay unless there is a way to overcome this problem.

asakura:
Where is to see the Benchmark task collected data points?

mtrudel:
Yeah, they're definitely an imperfect instrument. I look mostly at the 'requests per second' graphs for the request type under consideration, and then see if the difference is both 1. material (more than 5-10% typically), and 2. mostly one-sided. In this case:

image

you'll notice that we have about a 3-5% boost that's basically across the board. This is good!

asakura:
Great! Sounds like a success, right?

mtrudel:
I think so! I'm generally adopting a perf approach of compounding small wins. There have (and continue to be!) a bunch of small perf wins that are hiding in plain sight. Thanks for finding (and fixing!) one! I'll have a proper review done by afternoon UTC tomorrow.

mtrudel:
Sorry for the delay; this looks great!

PR #162, opened 2023-06-12
Better handle errors from Thousand Island when unable to get local/peer info

mtrudel:
basred on @asakura's [recent work](https://github.com/mtrudel/thousand_island/pull/62) on Thousand Island, we should be smarter about handling errors when asking for information about local or peer info. Sadly, there's not much we can do in the case where we DO get errors back, but at least we're now erroring with something more clear than a match error.

### DIFF
--- a/lib/bandit/http1/handler.ex
+++ b/lib/bandit/http1/handler.ex
@@ -63,8 +63,15 @@ defmodule Bandit.HTTP1.Handler do
   end
 
   defp build_transport_info(socket) do
-    {ThousandIsland.Socket.secure?(socket), ThousandIsland.Socket.local_info(socket),
-     ThousandIsland.Socket.peer_info(socket), ThousandIsland.Socket.telemetry_span(socket)}
+    secure? = ThousandIsland.Socket.secure?(socket)
+    telemetry_span = ThousandIsland.Socket.telemetry_span(socket)
+
+    with local_info when is_map(local_info) <- ThousandIsland.Socket.local_info(socket),
+         peer_info when is_map(peer_info) <- ThousandIsland.Socket.peer_info(socket) do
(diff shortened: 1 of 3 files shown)


lib/bandit/http1/handler.ex
```diff
@@ -63,8 +63,15 @@ defmodule Bandit.HTTP1.Handler do
   end
 
   defp build_transport_info(socket) do
-    {ThousandIsland.Socket.secure?(socket), ThousandIsland.Socket.local_info(socket),
-     ThousandIsland.Socket.peer_info(socket), ThousandIsland.Socket.telemetry_span(socket)}
+    secure? = ThousandIsland.Socket.secure?(socket)
+    telemetry_span = ThousandIsland.Socket.telemetry_span(socket)
+
+    with local_info when is_map(local_info) <- ThousandIsland.Socket.local_info(socket),
```
asakura, line 69: Should we keep local_info/1 and peer_info/1 return signatures closer to their counterparts from OTP library? Is so, I could do that later this week.

mtrudel: In terms of the map structures, you mean? 

I'd originally tried to cut them in the shape of the `Plug.Conn.Adapter.get_peer_data` call (and had managed to push those same semantics all the way down to the socket calls), but now that that's been forgone by the case block in this PR, I don't have any problem going the other way and pulling the signatures up from OTP. 

Or did you mean something else entirely? :)

asakura: That's exactly what I was wondering about :smiley:

mtrudel: Great. Let's get this landed and we can work that up as a separate PR. 

To that end, care to review what's here?

PR #163, opened 2023-06-12
Add typespecs to `Headers` module and remove unnecessary guard

ryanwinchester:
### Added
- added typespecs for the functions added in #148
- added tests for `get_content_length/1`(won't pass until we have a version with this PR https://github.com/elixir-plug/plug/pull/1155)

### Changed
- made `parse_integer/1` private

### Removed
- removed an unnecessary guard since the parser can never return a negative integer now
- removed tests for `parse_integer/1` (replaced by tests for get_content_length/1`)

## IMPORTANT NOTE

It's easy to miss if skimming: the updated tests **won't pass** _until_ we have a version of `plug` with the above-mentioned PR.

### DIFF
--- a/lib/bandit/headers.ex
+++ b/lib/bandit/headers.ex
@@ -15,7 +15,7 @@ defmodule Bandit.Headers do
 
   # Covers IPv6 addresses, like `[::1]:4000` as defined in RFC3986.
   @spec parse_hostlike_header(host_header :: binary()) ::
-          {:ok, binary(), nil | integer()} | {:error, String.t()}
+          {:ok, Plug.Conn.host(), nil | Plug.Conn.port_number()} | {:error, String.t()}
   def parse_hostlike_header("[" <> _ = host_header) do
     host_header
     |> :binary.split("]:")
@@ -46,7 +46,8 @@ defmodule Bandit.Headers do
     end
   end
 
-  @spec get_content_length(Plug.Conn.headers()) :: {:ok, nil | integer()} | {:error, String.t()}
+  @spec get_content_length(Plug.Conn.headers()) ::
+          {:ok, nil | non_neg_integer()} | {:error, String.t()}
   def get_content_length(headers) do
     case get_header(headers, "content-length") do
       nil -> {:ok, nil}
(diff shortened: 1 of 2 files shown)


lib/bandit/headers.ex
```diff
@@ -46,22 +46,23 @@ defmodule Bandit.Headers do
...
       value -> parse_content_length(value)
     end
   end
 
-  @spec parse_content_length(binary()) :: {:ok, length :: integer()} | {:error, String.t()}
+  @spec parse_content_length(binary()) :: {:ok, non_neg_integer()} | {:error, String.t()}
   defp parse_content_length(value) do
     case parse_integer(value) do
-      {length, ""} when length >= 0 ->
+      {length, ""} ->
```
ryanwinchester, line 61: It will _always_ be a non-negative integer or `:error`

lib/bandit/headers.ex
```diff
@@ -46,22 +46,23 @@ defmodule Bandit.Headers do
...
         {:ok, length}
 
-      {length, rest} ->
-        if rest |> Plug.Conn.Utils.list() |> all?(to_string(length)),
+      {length, _rest} ->
+        if value |> Plug.Conn.Utils.list() |> all?(to_string(length)),
```
ryanwinchester, line 65: Previously this had a bug where it would still work for an invalid content-length header like `123 123`

ryanwinchester:
> @ryanwinchester can you hold off on this for a day or so until #165 gets over the line?

Yeah we need to wait for a Plug release anyway.

mtrudel:
> @mtrudel Yeah, we need to wait for a Plug release anyway.

@ryanwinchester if I understand this correctly there's no runtime concern to people still using older version of plug with this merged, correct? It's just a concern for tests?

ryanwinchester:
Good here.

Notes:

* dialyzer errors in `thousand_island` will be resolved when we have https://github.com/mtrudel/thousand_island/commit/50447270eba5676597a5ac727a935cbfe3a34909
* I skipped the failing test case and we can unskip when Plug updates.

mtrudel:
> dialyzer errors from thousand_island will be resolved when we have https://github.com/mtrudel/thousand_island/commit/50447270eba5676597a5ac727a935cbfe3a34909

This is in main now - can you rebase and push up so we can see green CI here?

ryanwinchester:
👍

mtrudel:
Thanks for the PR @ryanwinchester !

PR #165, opened 2023-06-12
Improve handling of Plug-provided content headers

mtrudel:
Revise behaviour to now:

1. Not attempt to content-encode a response if the Plug has already set a `content-encoding` header
2. Overwrite the content-length provided by Plug in all cases (this has the side effect of now including `content-length` headers in HTTP/2 responses)

### DIFF
--- a/lib/bandit/http2/adapter.ex
+++ b/lib/bandit/http2/adapter.ex
@@ -124,8 +124,11 @@ defmodule Bandit.HTTP2.Adapter do
           {body, headers, %{}}
       end
 
+    body_bytes = IO.iodata_length(body)
+    headers = Bandit.Headers.add_content_length(headers, body_bytes, status)
+
     adapter =
-      if IO.iodata_length(body) == 0 do
+      if body_bytes == 0 do
         adapter
         |> send_headers(status, headers, true)
       else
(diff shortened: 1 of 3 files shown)


lib/bandit/headers.ex
```diff
@@ -61,7 +61,7 @@ defmodule Bandit.Headers do
         {:ok, length}
 
       {length, rest} ->
-        if rest |> Plug.Conn.Utils.list() |> all?(to_string(length)),
+        if rest |> Plug.Conn.Utils.list() |> Enum.all?(&(&1 == to_string(length))),
```
ryanwinchester, line 64: 💯

lib/bandit/headers.ex
```diff
@@ -89,4 +83,18 @@ defmodule Bandit.Headers do
   end
 
   defp parse_integer(rest, total), do: {total, rest}
+
+  def add_content_length(headers, length, status) do
+    headers = Enum.reject(headers, &(elem(&1, 0) == "content-length"))
```
ryanwinchester, line 88: Maybe

```suggestion
    headers = List.keydelete(headers, "content-length", 0)
```

mtrudel: Alas, I want to delete *all* possible occurrences of content-length in the input header list, and `keydelete/3` only deletes the first.

In general I find the `List` functions inherited from `:list` to be super awkward to use (like, for example, `keyreplace` will only replace, not upsert). They're an oddball interface that I've never found the right place to make heavy use of.

PR #166, opened 2023-06-13
Reduce amount of work HTTP1 Adapter's do to ready body chunks

asakura:
It was difficult to measure impact of the changes provided, but I hope that 
* the compiler can inline specialized version of the read function which would lead to less instructions executed in runtime
* passing around refs to variables faster than changing content of a map

### DIFF
--- a/lib/bandit/http1/adapter.ex
+++ b/lib/bandit/http1/adapter.ex
@@ -18,6 +18,24 @@ defmodule Bandit.HTTP1.Adapter do
             websocket_enabled: false,
             opts: []
 
+  @type t :: %__MODULE__{
+          state: state(),
+          socket: ThousandIsland.Socket.t(),
+          buffer: binary(),
+          body_remaining: nil | integer(),
+          body_encoding: nil | binary(),
+          version: nil | :"HTTP/1.1" | :"HTTP/1.0",
+          keepalive: boolean(),
+          content_encoding: String.t(),
+          upgrade: nil | {:websocket, opts :: keyword(), websocket_opts :: keyword()},
+          metrics: %{},
+          websocket_enabled: boolean(),
+          opts: %{
+            required(:http_1) => Bandit.http_1_options(),
+            required(:websocket) => Bandit.websocket_options()
+          }
+        }
+
   ################
   # Header Reading
   ################
@@ -75,9 +93,9 @@ defmodule Bandit.HTTP1.Adapter do
 
     case :erlang.decode_packet(type, req.buffer, packet_size: packet_size) do
       {:more, _len} ->
-        with {:ok, iodata} <- read(req.socket, 0) do
+        with {:ok, chunk} <- read_available(req.socket, _read_timeout = nil) do
           # decode_packet expects a binary, so convert it to one
-          req = %{req | buffer: IO.iodata_to_binary([req.buffer | iodata])}
+          req = %{req | buffer: req.buffer <> chunk}
           do_read_headers(req, type, headers, method, request_target)
         end
 
@@ -158,7 +176,10 @@ defmodule Bandit.HTTP1.Adapter do
   # Body Reading
   ##############
 
+  @dialyzer {:no_improper_lists, read_req_body: 2}
   @impl Plug.Conn.Adapter
+  @spec read_req_body(t(), keyword()) ::
+          {:ok, data :: binary(), t()} | {:more, data :: binary(), t()} | {:error, term()}
   def read_req_body(%__MODULE__{state: :no_body} = req, _opts) do
     time = Bandit.Telemetry.monotonic_time()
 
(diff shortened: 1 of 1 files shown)


lib/bandit/http1/adapter.ex
```diff
@@ -75,9 +93,9 @@ defmodule Bandit.HTTP1.Adapter do
 
     case :erlang.decode_packet(type, req.buffer, packet_size: packet_size) do
       {:more, _len} ->
-        with {:ok, iodata} <- read(req.socket, 0) do
+        with {:ok, chunk} <- read_available(req.socket, _read_timeout = nil) do
           # decode_packet expects a binary, so convert it to one
-          req = %{req | buffer: IO.iodata_to_binary([req.buffer | iodata])}
+          req = %{req | buffer: req.buffer <> chunk}
```
ryanwinchester, line 98: Why the concat?

asakura: Because `chunk` is guaranteed to be a binary here.

ryanwinchester: Okay so assume (correctly) that I don't know why that makes it better.

As my understanding goes: string concat is O(n) and adding to improper list is O(1).
I know big-O by itself isn't always the best way to compare performance and there are subtleties if you know ahead of time what your data looks like, so I was hoping you could explain why concat would be better here. Is it just because you use `byte_size/1` later which is better than `IO.iodata_length/1`? And if so why the swap? Or is there more to it?

asakura: You mean that snippet?

```elixir
        result = IO.iodata_to_binary([buffer | iolist])
        result_size = byte_size(result)
        body_remaining = body_remaining - result_size - byte_size(buffer)
```

`byte_size/1` executes in constant time irrelevant to the size of the binary, while `IO.iodata_length/1` has to traverse entire the iodata list for the second time (the first time was when `IO.iodata_to_binary/1` converted the iodata to a binary and stored its size).

```elixir
          req = %{req | buffer: IO.iodata_to_binary([req.buffer | iodata])}
          req = %{req | buffer: req.buffer <> chunk}
```

Yes, indeed - appending to improper list is O(1) + an extra word allocation and if that data would go straight over the wire, that would be all complexity. But `req.buffer` is a binary and freshly constructed list need to be converted to a binary with `IO.iodata_to_binary/1`. What is complexity of `IO.iodata_to_binary/1`? 

On the other size `<>/2` compiles to bit syntax like `<<Buffer/binary, Chunk/binary>>` and according to the documentation (https://www.erlang.org/doc/efficiency_guide/binaryhandling.html#constructing-binaries) BEAM might enable an optimization which would prevent those binaries from being copied by placing `Chunk` in to `Buffer`.

ryanwinchester: Thanks and yeah this line of code just clicked to me why it is obviously better.

asakura: https://github.com/erlang/otp/blob/maint-26/erts/emulator/beam/binary.c#L928 Erlang's implementation of iolist_to_binary BIF. Sorry, I don't have much time at the moment to study it.

ryanwinchester: I need to stop reading PRs in the morning before coffee.

lib/bandit/http1/adapter.ex
```diff
@@ -286,30 +326,27 @@ defmodule Bandit.HTTP1.Adapter do
...
-        ) :: {:ok, iolist()} | {:error, :closed | :timeout | :inet.posix()}
-  defp read(socket, to_read, opts \\ []) do
-    read_size = Keyword.get(opts, :read_length, 1_000_000)
-    read_timeout = Keyword.get(opts, :read_timeout)
-    do_read(socket, to_read, [], read_size, read_timeout)
+  @compile {:inline, read_available: 2}
+  @spec read_available(ThousandIsland.Socket.t(), timeout()) ::
+          {:ok, binary()} | {:error, :closed | :timeout | :inet.posix()}
+  defp read_available(socket, read_timeout) do
+    ThousandIsland.Socket.recv(socket, 0, read_timeout)
```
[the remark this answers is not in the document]
asakura: This patch does not change the code logic at all - it works the same way it worked before. The only thing which differs is that now it's obvious that `ThousandIsland.Socket.recv(socket, 0, read_timeout)` is invoked in some cases (I've just spotted that `read_timeout` is `nil` in some cases otherwise whatever user provided in `Keyword.get(opts, :read_timeout)`). Before that was hidden in `read/3` function.

Unfortunately the documentation isn't clear on how `0` works with a timeout. But it says that timeout's default values is `:infinity` (not `nil` which we have at the moment, I think it would be beneficial to fix that right away):
```
Receives a packet from a socket in passive mode. A closed socket is indicated by return value {error, closed}.

Argument Length is only meaningful when the socket is in raw mode and denotes the number of bytes to read. If Length is 0, all available bytes are returned. If Length > 0, exactly Length bytes are returned, or an error; possibly discarding less than Length bytes of data when the socket is closed from the other side.

The optional Timeout parameter specifies a time-out in milliseconds. Defaults to infinity.
```

What do you think if we move discussion on these concerns to a new issue and eventually work out a separate PR?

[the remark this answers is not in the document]
mtrudel: Good point. I think this reads a lot better overall, given that it's the same net effect.

Agreed that we should fix that everywhere on a separate PR I think.

PR #167, opened 2023-06-14
Adopt renamed functions from ThousandIsland.Socket module

asakura:


lib/bandit/http1/adapter.ex
```diff
@@ -503,14 +504,57 @@ defmodule Bandit.HTTP1.Adapter do
...
+  def to_plug_peer_data({:local, path}, cert),
+    do: %{address: {:local, path}, port: 0, ssl_cert: cert}
+
+  def to_plug_peer_data({:unspec, <<>>}, cert),
+    do: %{address: :unspec, port: 0, ssl_cert: cert}
+
+  def to_plug_peer_data({:undefined, term}, cert),
+    do: %{address: {:undefined, term}, port: 0, ssl_cert: cert}
+
+  def to_plug_peer_data({ip, port}, cert), do: %{address: ip, port: port, ssl_cert: cert}
```
mtrudel, line 559: These should be `defp`?

asakura: Yes, it should be!

PR #168, opened 2023-06-16
Remove comment that no longer makes sense

ryanwinchester:
Makes no sense since this was changed a few days ago.

### DIFF
--- a/lib/bandit/http1/adapter.ex
+++ b/lib/bandit/http1/adapter.ex
@@ -94,7 +94,6 @@ defmodule Bandit.HTTP1.Adapter do
     case :erlang.decode_packet(type, req.buffer, packet_size: packet_size) do
       {:more, _len} ->
         with {:ok, chunk} <- read_available(req.socket, _read_timeout = nil) do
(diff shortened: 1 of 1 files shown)


mtrudel:
I added that comment when I was working up an 'iolist all the things!' push to remind the reader that there's a reason we're not just making an iolist here. 

To that end, I think this still makes sense, since we're recursing on `do_read_headers` and it directly calls decode_packet.  Maybe we could reword it?

ryanwinchester:
It was there from [when you were converting it to a binary](https://github.com/mtrudel/bandit/commit/08e06d3eac59c3a265c52d769bc327c605f3ca94#diff-e4d9e1fa460c65a2913911c622ecc4963e47401ff7b46515627374bed49669c9R97-R98) with `IO.iodata_to_binary/1`, though. 

https://github.com/mtrudel/bandit/blob/9994870b7a9aa7b56cffd44228a53a3cbcfaed6c/lib/bandit/http1/adapter.ex#L78-L82

mtrudel:
Yeah I get it! Cool. Merging.

PR #170, opened 2023-06-20
Fix failing connection when peer doesn't provide a certificate

moogle19:
Currently all HTTPS connection via HTTP1 will fail, because a client certificate is required (?) when using TLS.
This changes the behavior to allow HTTPS connections without a client certificate

### DIFF
--- a/lib/bandit/http1/adapter.ex
+++ b/lib/bandit/http1/adapter.ex
@@ -510,6 +510,9 @@ defmodule Bandit.HTTP1.Adapter do
         {:ok, cert} ->
           cert
 
+        {:error, :no_peercert} ->
+          nil
+
         {:error, reason} ->
           raise "Unable to obtain peer cert: #{inspect(reason)}"
       end


moogle19:
Example error when making a HTTP1 HTTPS request:

```elixir
[error] GenServer #PID<0.2323.0> terminating
** (RuntimeError) Unable to obtain peer cert: :no_peercert
    (bandit 1.0.0-pre.9) lib/bandit/http1/adapter.ex:515: Bandit.HTTP1.Adapter.get_peer_cert!/1
    (bandit 1.0.0-pre.9) lib/bandit/http1/adapter.ex:527: Bandit.HTTP1.Adapter.get_peer_data/1
```

asakura:
So the socket is secure but it still for some reason cannot retrieve peer certificate? How come?
Could find that thread from 2012 about similar issue: http://erlang.org/pipermail/erlang-questions/2012-November/070736.html
And there is a issue in https://github.com/erlang/otp/issues/4800

moogle19:
Normally a client has no peer certificate, except when it is using Two-Way authentication. But that is quite rare.

asakura:
Ah right. That's make sense now. We are providing Plug with `%{address: ..., port: .., ssl_cert: cert}` all the time (even in cases when  `{:verify, :verify_peer}`  is not an option for the transport).

mtrudel:
Just dug into this to try and get some test coverage on the original issue, and this wouldn't have actually broken *all* HTTP/1 connections over SSL, just those that made use of the `get_peer_data/1` function. We added coverage for that at https://github.com/mtrudel/bandit/pull/178/files#diff-6a1dfdf6de8e0ba81a6ae4eef3f45cf79b7d4a55d7130e228463f9f77a0344e3R1027-R1040 so I think we're all good for future regressions here.

PR #171, opened 2023-06-20
Fix peer_info for HTTP2 and unify behavior

moogle19:
Currently the `Plug.Conn.get_peer_data/1` function returns the wrong format in HTTP2 requests (it returns `ThousandIsland.Transport.on_peername()` instead of `Plug.Conn.Adapter.peer_data()`). 

This fixes the behavior and unifies it between HTTP1 and HTTP2 at the cost of trying to fetch the `peer_cert` for every request (for HTTP1).

~~There is also the duplication of `get_peer_cert!` in `HTTP1.Handler` and `HTTP2.Connection`. I would like to move it to a separate module, but couldn't find the right one yet.~~

The unification also resulted in a new module `Bandit.Plug`. Not sure about the name and possible confusion. I am open for renaming suggestions.

Edit:
I removed `get_local_data/1` since it wasn't used (?) and it is not part of the `Plug` callbacks.


moogle19:
Example error message from Sentry Plug:

```elixir
[error] Task #PID<0.2300.0> started from #PID<0.2299.0> terminating
** (KeyError) key :port not found in: {{127, 0, 0, 1}, 65342}

If you are using the dot syntax, such as map.field, make sure the left-hand side of the dot is a map
    (sentry 8.0.6) lib/sentry/plug_context.ex:127: Sentry.PlugContext.build_request_interface_data/2
    (sentry 8.0.6) lib/sentry/plug_context.ex:99: Sentry.PlugContext.call/2
```

mtrudel:
Obsoleted by #178!

PR #172, opened 2023-06-20
It should not subtract previous buffer size from body_remaining

asakura:


lib/bandit/http1/adapter.ex
```diff
@@ -234,7 +234,7 @@ defmodule Bandit.HTTP1.Adapter do
       with {:ok, iolist} <- read(req.socket, to_read, [], read_size, read_timeout) do
         result = IO.iodata_to_binary([buffer | iolist])
         result_size = byte_size(result)
-        body_remaining = body_remaining - result_size - byte_size(buffer)
+        body_remaining = body_remaining - result_size + byte_size(buffer)
```
mtrudel, line 237: Shouldn't this be 
```suggestion
        body_remaining = body_remaining - result_size
```

`result` already includes the content of buffer (from line 235), so isn't this double counting the value in buffer in both the `+` and `-` cases?

asakura: image

In the original version it was subtracting iodata length (iolist in this version) from the body_remaining. Well, now I realize that I don't understand why it was like that. From logical perspective of view it seems right to subtract the entry size from the body_remaining variable.

mtrudel: Are you saying it was wrong before, and is now wrong in a different way?

asakura: So apparently it was wrong before and now is wrong in a different way :laughing:

PR #174, opened 2023-06-21
Fix handling of requests with more content than given in content-length

moogle19:
Currently if a request has a larger body than given in `content-length` we get the error:
```elixir
[error] GenServer #PID<0.720.0> terminating
** (FunctionClauseError) no function clause matching in :prim_inet.recv0/3
    :prim_inet.recv0(#Port<0.22>, -25, 60000)
    (thousand_island 1.0.0-pre.5) lib/thousand_island/socket.ex:59: ThousandIsland.Socket.recv/3
    (bandit 1.0.0-pre.9) lib/bandit/http1/adapter.ex:346: Bandit.HTTP1.Adapter.read/5
    (bandit 1.0.0-pre.9) lib/bandit/http1/adapter.ex:234: Bandit.HTTP1.Adapter.read_req_body/2
    (plug 1.14.2) lib/plug/conn.ex:1153: Plug.Conn.read_body/2
    (phx_test 0.1.0) lib/phx_test_web/controllers/some_controller.ex:11: PhxTestWeb.SomeController.a/2
    (phx_test 0.1.0) lib/phx_test_web/controllers/some_controller.ex:1: PhxTestWeb.SomeController.action/2
    (phx_test 0.1.0) lib/phx_test_web/controllers/some_controller.ex:1: PhxTestWeb.SomeController.phoenix_controller_pipeline/2
    (phoenix 1.7.6) lib/phoenix/router.ex:430: Phoenix.Router.__call__/5
    (phx_test 0.1.0) lib/phx_test_web/endpoint.ex:1: PhxTestWeb.Endpoint.plug_builder_call/2
    (phx_test 0.1.0) deps/plug/lib/plug/debugger.ex:136: PhxTestWeb.Endpoint."call (overridable 3)"/2
    (phx_test 0.1.0) lib/phx_test_web/endpoint.ex:1: PhxTestWeb.Endpoint.call/2
    (phoenix 1.7.6) lib/phoenix/endpoint/sync_code_reload_plug.ex:22: Phoenix.Endpoint.SyncCodeReloadPlug.do_call/4
    (bandit 1.0.0-pre.9) lib/bandit/pipeline.ex:110: Bandit.Pipeline.call_plug/2
    (bandit 1.0.0-pre.9) lib/bandit/pipeline.ex:25: Bandit.Pipeline.run/6
    (bandit 1.0.0-pre.9) lib/bandit/http1/handler.ex:27: Bandit.HTTP1.Handler.handle_data/3
    (bandit 1.0.0-pre.9) lib/bandit/delegating_handler.ex:18: Bandit.DelegatingHandler.handle_data/3
    (bandit 1.0.0-pre.9) /Users/kevin/Developer/repos/phx_test/deps/thousand_island/lib/thousand_island/handler.ex:399: Bandit.DelegatingHandler.handle_continue/2
    (stdlib 5.0.1) gen_server.erl:1067: :gen_server.try_handle_continue/3
    (stdlib 5.0.1) gen_server.erl:977: :gen_server.loop/7
State: {%ThousandIsland.Socket{socket: #Port<0.22>, transport_module: ThousandIsland.Transports.TCP, read_timeout: 60000, span: %ThousandIsland.Telemetry{span_name: :connection, telemetry_span_context: #Reference<0.513344876.2758541319.164812>, start_time: -576460742372031167, start_metadata: %{remote_address: {127, 0, 0, 1}, remote_port: 51244, telemetry_span_context: #Reference<0.513344876.2758541319.164812>, parent_telemetry_span_context: #Reference<0.513344876.2758541316.161650>}}}, %{opts: %{websocket: [], http_1: [], http_2: []}, plug: {Phoenix.Endpoint.SyncCodeReloadPlug, {PhxTestWeb.Endpoint, []}}, websocket_enabled: true, handler_module: Bandit.InitialHandler, http_1_enabled: true, http_2_enabled: true}}
```

because [this](https://github.com/mtrudel/bandit/blob/5268c2eb5e8a86c0583b815c9918b829e0a88d88/lib/bandit/http1/adapter.ex#L64) will get negative, since we do not limit the first read.

This adds a handler for negative values and only returns the amount of bytes given in `content-length` as body.

### DIFF
--- a/lib/bandit/http1/adapter.ex
+++ b/lib/bandit/http1/adapter.ex
@@ -193,37 +193,28 @@ defmodule Bandit.HTTP1.Adapter do
   end
 
   def read_req_body(
-        %__MODULE__{state: :headers_read, buffer: buffer, body_remaining: 0} = req,
-        _opts
-      ) do
-    time = Bandit.Telemetry.monotonic_time()
-
-    metrics =
-      req.metrics
-      |> Map.update(:req_body_bytes, byte_size(buffer), &(&1 + byte_size(buffer)))
-      |> Map.put_new(:req_body_start_time, time)
-      |> Map.put(:req_body_end_time, time)
-
-    {:ok, buffer, %{req | state: :body_read, buffer: <<>>, metrics: metrics}}
-  end
-
-  def read_req_body(
         %__MODULE__{state: :headers_read, buffer: buffer, body_remaining: body_remaining} = req,
         _opts
       )
-      when body_remaining < 0 do
-    buffer_size = byte_size(buffer)
-    # Since we read to much data and body_remaining is negative, we can add it to the buffer_size
-    # to get the content-length of the body
-    binary_size = buffer_size + body_remaining
+      when body_remaining <= 0 do
+    {body, body_size} =
+      if body_remaining == 0 do
+        {buffer, byte_size(buffer)}
+      else
+        body_size = byte_size(buffer) + body_remaining
 
-    <<body::binary-size(binary_size), _rest::bitstring>> = buffer
+        # Since we read to much data and body_remaining is negative, we can add it to the buffer_size
+        # to get the content-length of the body
+        <<body::binary-size(body_size), _rest::bitstring>> = buffer
+
+        {body, body_size}
+      end
 
     time = Bandit.Telemetry.monotonic_time()
 
     metrics =
       req.metrics
-      |> Map.update(:req_body_bytes, binary_size, &(&1 + binary_size))
+      |> Map.update(:req_body_bytes, body_size, &(&1 + body_size))
       |> Map.put_new(:req_body_start_time, time)
       |> Map.put(:req_body_end_time, time)
 
--- a/lib/bandit/http1/adapter.ex
+++ b/lib/bandit/http1/adapter.ex
@@ -208,6 +208,29 @@ defmodule Bandit.HTTP1.Adapter do
   end
 
   def read_req_body(
+        %__MODULE__{state: :headers_read, buffer: buffer, body_remaining: body_remaining} = req,
+        _opts
+      )
+      when body_remaining < 0 do
+    buffer_size = byte_size(buffer)
+    # Since we read to much data and body_remaining is negative, we can add it to the buffer_size
+    # to get the content-length of the body
+    binary_size = buffer_size + body_remaining
+
+    <<body::binary-size(binary_size), _rest::bitstring>> = buffer
+
+    time = Bandit.Telemetry.monotonic_time()
+
+    metrics =
+      req.metrics
+      |> Map.update(:req_body_bytes, binary_size, &(&1 + binary_size))
+      |> Map.put_new(:req_body_start_time, time)
+      |> Map.put(:req_body_end_time, time)
+
+    {:ok, body, %{req | state: :body_read, buffer: <<>>, metrics: metrics}}
+  end
+
+  def read_req_body(
         %__MODULE__{state: :headers_read, body_remaining: body_remaining, buffer: buffer} = req,
         opts
       )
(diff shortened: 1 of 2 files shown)


asakura:
Hi @moogle19 

I'm afraid that's happened because line 260 should be `body_remaining = body_remaining - result_size + byte_size(buffer)`. Could you try that please?

moogle19:
@asakura 
No that is already to late.

This is called before:
https://github.com/mtrudel/bandit/blob/5268c2eb5e8a86c0583b815c9918b829e0a88d88/lib/bandit/http1/adapter.ex#L232-L235

where `to_read` will be negative, since `body_remaining` is negative and we call `min/2` there.

asakura:
Clear. `read_available/2` might be called one or more times in `do_read_headers/5`. That would make it to return `req` with more data in `req.buffer` than given in `content-length`. 

Should we just discard those extra bytes after chopping `req.buffer` or keep them in `req.buffer`? And why would someone send more data than they would announce in `content-length`? Is that HTTP Pipelining?

moogle19:
> Should we just discard those extra bytes after chopping req.buffer or keep them in req.buffer?

I wasn't quite sure about keeping the additional data in the buffer or not. Maybe @mtrudel could help here.

> And why would someone send more data than they would announce in content-length? Is that HTTP Pipelining?

Not sure why someone would do it. I just noticed it while testing the broken multipart requests and thought it should be fixed.

asakura:
I think @mtrudel comments is related here https://github.com/mtrudel/bandit/pull/166#discussion_r1228125555

mtrudel:
I think the best path forward here, rather than trying to reason about all of these different cases, is to get some tests covering the various 'bytes available vs requested' scenarios. As you're both well aware, this stuff is *incredibly* tricky, and fixing one problem often creates another. I was going to write such tests up on the heels of the bug fixed in #172, and seeing this one hanging out just increases the need for them IMO.

mtrudel:
There's some pretty comprehensive refactoring / test coverage at #184 that will make this a whole lot easier to reason about. Let's revisit this once that lands

mtrudel:
@moogle19 I'd started down the same path you did here, thinking that if the client sends more than the content length says they should that we should just discard the rest (this was even the spirit of the skipped test I'd put up in #184 that I was thinking we'd handle here). However, I think that this is wrong for a couple of reasons:

1. A properly behaved client should never do this. We should be surfacing this misbehaviour somehow
2. It's also perilously close to request smuggling, and we ought to make that very apparent to the user

I've pushed up a branch that handle this at https://github.com/mtrudel/bandit/tree/mt_handle_too_large_bodies if you want to crib from it. I don't want to mess with your PR too much considering it's branched from before #184 landed.

LMK what you think!

moogle19:
@mtrudel 
Do we want to support HTTP Pipelining in the future? 
If not, I think we can use your version in `mt_handle_too_large_bodies`.

Since HTTP Pipelining is disabled in all major browsers, I think it is okay if we don't support it.

(I tested HTTP Pipelining with Cowboy and Go (at least how I think it is working 😆  ):
Cowboy seem to support it, but only if the requests are not transmitted as a whole.
Go supports HTTP Pipelining even when 2 requests are submitted as a whole)

This works for Go, but not for Cowboy:
```bash
nc -c localhost 8080 <<EOF
GET / HTTP/1.1
Host: localhost:8080
Content-Length: 0
User-Agent: curl/7.54.0
Accept: */*

GET / HTTP/1.1
Host: localhost:8080
Content-Length: 0
User-Agent: curl/7.54.0
Accept: */*

EOF
```

doing 
```bash
nc -c localhost 8080
```
and than pasting the 2 GET-Requests in there works for Cowboy and Go.
Bandit discards the second request.

mtrudel:
Pipelining is fundamentally incompatible with the Bandit process model of 'one process per connection'. This is one place where Cowboy's 'one process per connection and one per request within the connection' process model is more flexible. However, I deliberately don't want to go down that road as 1. having half the process count for HTTP/1 requests is a huge win, and 2. pipelining is really tricky to pull off for almost zero benefit.

All of which is to say that I'm not losing any sleep about not having support for it.

mtrudel:
Closed in favour of #190

PR #183, opened 2023-06-23
Compensate remaining body length by previous buffer size

asakura:
As shown in https://github.com/mtrudel/bandit/pull/182, `body_remaining` goes down beyond 0. That signalizes that due to size miscalculation, `do_req_body` function skips on reading the last few packets from the socket which contains the tail of the body. Note that `body_remaining` is already being adjusted on line 62 after the headers were read. I'm certain that this should not happen. I'm not sure if that fixes `BodyAlreadyReadError` but I think it's still worth to have that fix in the main branch :)


mtrudel:
I literally just discovered the *exact* same thing, though I'm solving it a smidge differently (DRYing up a few other byte_size calls in that same function). 

PR up momentarily.

mtrudel:
Closed in favour of #184

Issue #169, opened 2023-06-19
Bandit.BodyAlreadyReadError error raising since upgrade to 1.0.0-pre9


scollon42:
Hi there!

First I wanted to thank you all for your amazing work! I really enjoyed using your library with Phoenix so far 👌 💜

 I'm very sorry, I will try to give as much information as I can but I'm failing to find what is happening exactly here. 🤔 

I've upgraded to bandit 1.0.0-pre9 today (i was in pre6 before) and a new error appeared.

It's an endpoint that has not been updated and Bandit is raising a `BodyAlreadyReadError` 🤔 

I have faily small stacktrace to share : 
```
Plug.Parsers.ParseError: malformed request, a Bandit.BodyAlreadyReadError exception was raised with message "Body has already been read"
  File "lib/bandit/http1/adapter.ex", line 288, in Bandit.HTTP1.Adapter.read_req_body/2
  File "lib/plug/conn.ex", line 1290, in Plug.Conn.next_multipart/3
  File "lib/plug/conn.ex", line 1255, in Plug.Conn.read_part_body/8
  File "lib/plug/parsers/multipart.ex", line 247, in Plug.Parsers.MULTIPART.parse_multipart_file/4
  File "lib/plug/parsers/multipart.ex", line 215, in Plug.Parsers.MULTIPART.parse_multipart_headers/5
  File "lib/plug/parsers/multipart.ex", line 186, in Plug.Parsers.MULTIPART.parse_multipart/5
  File "lib/plug/parsers/multipart.ex", line 175, in Plug.Parsers.MULTIPART.parse_multipart/2
  File "lib/plug/parsers/multipart.ex", line 127, in Plug.Parsers.MULTIPART.parse/5
```

On the other way. Requests seems to still go through. (This endpoint in particular is a webhook handler, and events are seems to be processed properly).
The endpoint is fairly simple, there is no plug or anything.
Also it doesn't happened with every call to this endpoint. I haven't seen any differences between calls that failed and calls that succeed.

Again, I have trouble identifying from where it can source. I don't see changes in the codebase except for the version bump 🤔 

```ex
[...]

PR #184, opened 2023-06-23
Simplify content-length body reading

mtrudel:
Duplicates #183 
Fixes #182
Related to #169

### DIFF
--- a/lib/bandit/http1/adapter.ex
+++ b/lib/bandit/http1/adapter.ex
@@ -96,7 +96,7 @@ defmodule Bandit.HTTP1.Adapter do
 
     case :erlang.decode_packet(type, req.buffer, packet_size: packet_size) do
       {:more, _len} ->
-        with {:ok, chunk} <- read_available(req.socket, _read_timeout = nil) do
+        with {:ok, chunk} <- read_available(req.socket) do
           req = %{req | buffer: req.buffer <> chunk}
           do_read_headers(req, type, headers, method, request_target)
         end
@@ -194,37 +194,14 @@ defmodule Bandit.HTTP1.Adapter do
     {:ok, <<>>, %{req | metrics: metrics}}
   end
 
-  def read_req_body(
-        %__MODULE__{state: :headers_read, body_remaining: body_remaining, buffer: buffer} = req,
-        opts
-      )
+  def read_req_body(%__MODULE__{state: :headers_read, body_remaining: body_remaining} = req, opts)
       when is_number(body_remaining) do
     metrics =
       Map.put_new_lazy(req.metrics, :req_body_start_time, &Bandit.Telemetry.monotonic_time/0)
 
-    max_desired_bytes = Keyword.get(opts, :length, 8_000_000)
-
-    result =
-      if byte_size(buffer) >= max_desired_bytes || body_remaining == 0 do
-        # We can satisfy the read request entirely from our buffer
-        bytes_to_return = min(max_desired_bytes, byte_size(buffer))
-        <<to_return::binary-size(bytes_to_return), rest::binary>> = buffer
-        {:ok, to_return, rest, body_remaining}
-      else
-        # We need to read off the wire
-        bytes_to_read = min(max_desired_bytes - byte_size(buffer), body_remaining)
-        read_size = Keyword.get(opts, :read_length, 1_000_000)
-        read_timeout = Keyword.get(opts, :read_timeout)
-
-        with {:ok, iolist} <- read(req.socket, bytes_to_read, [], read_size, read_timeout) do
-          to_return = IO.iodata_to_binary([buffer | iolist])
-          body_remaining = body_remaining - IO.iodata_length(iolist)
-          {:ok, to_return, <<>>, body_remaining}
-        end
-      end
-
-    with {:ok, to_return, new_buffer, body_remaining} <- result do
-      if new_buffer == <<>> && body_remaining == 0 do
+    with {:ok, to_return, buffer, body_remaining} <-
+           do_read_content_length_body(req.socket, req.buffer, body_remaining, opts) do
+      if buffer == <<>> && body_remaining == 0 do
         metrics =
           metrics
           |> Map.update(:req_body_bytes, byte_size(to_return), &(&1 + byte_size(to_return)))
@@ -238,31 +215,21 @@ defmodule Bandit.HTTP1.Adapter do
           |> Map.update(:req_body_bytes, byte_size(to_return), &(&1 + byte_size(to_return)))
 
         {:more, to_return,
-         %{req | buffer: new_buffer, body_remaining: body_remaining, metrics: metrics}}
+         %{req | buffer: buffer, body_remaining: body_remaining, metrics: metrics}}
       end
     end
   end
 
-  def read_req_body(
-        %__MODULE__{
-          state: :headers_read,
-          body_encoding: "chunked",
-          socket: socket,
-          buffer: buffer
-        } = req,
-        opts
-      ) do
-    start_time = Bandit.Telemetry.monotonic_time()
-    read_size = Keyword.get(opts, :read_length, 1_000_000)
-    read_timeout = Keyword.get(opts, :read_timeout)
+  def read_req_body(%__MODULE__{state: :headers_read, body_encoding: "chunked"} = req, opts) do
+    metrics =
+      Map.put_new_lazy(req.metrics, :req_body_start_time, &Bandit.Telemetry.monotonic_time/0)
 
-    with {:ok, body, buffer} <- do_read_chunk(socket, buffer, <<>>, read_size, read_timeout) do
+    with {:ok, body, buffer} <- do_read_chunked_body(req.socket, req.buffer, <<>>, opts) do
       body = IO.iodata_to_binary(body)
 
       metrics =
-        req.metrics
+        metrics
         |> Map.put(:req_body_bytes, byte_size(body))
-        |> Map.put(:req_body_start_time, start_time)
         |> Map.put(:req_body_end_time, Bandit.Telemetry.monotonic_time())
 
       {:ok, body, %{req | buffer: buffer, metrics: metrics}}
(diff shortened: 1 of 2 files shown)


mtrudel:
@asakura check it out. I started with the same realization you did on #183, and then realized that the `body_remaining: 0` function clause could disappear entirely if it collapsed into the primary content-length `read_req_body` clause. Then I realized that I could de-dupe a bunch of the return handling by moving that out to a trailing `with` block. 

I think it reads a whole lot cleaner than any version we've ever had. WDYT?

lib/bandit/http1/adapter.ex
```diff
@@ -288,8 +243,29 @@ defmodule Bandit.HTTP1.Adapter do
...
+      bytes_to_return = min(max_desired_bytes, byte_size(buffer))
+      <<to_return::binary-size(bytes_to_return), rest::binary>> = buffer
+      {:ok, to_return, rest, body_remaining}
+    else
+      # We need to read off the wire
+      bytes_to_read = min(max_desired_bytes - byte_size(buffer), body_remaining)
+
+      with {:ok, iolist} <- read(socket, bytes_to_read, [], opts) do
+        to_return = IO.iodata_to_binary([buffer | iolist])
+        body_remaining = body_remaining - IO.iodata_length(iolist)
```
asakura, line 261: Just keep in mind that `IO.iodata_to_binary` uses the same function under the hood which `IO.iodata_length` calls. So basically the same job is done here twice at cost of extra reduction which will be deducted from a process's budget running this adapter.

lib/bandit/http1/adapter.ex
```diff
@@ -194,87 +194,42 @@ defmodule Bandit.HTTP1.Adapter do
...
-        result_size = byte_size(result)
-        body_remaining = body_remaining - result_size
+    metrics =
+      Map.put_new_lazy(req.metrics, :req_body_start_time, &Bandit.Telemetry.monotonic_time/0)
 
-        if body_remaining > 0 do
-          metrics = Map.update(metrics, :req_body_bytes, result_size, &(&1 + result_size))
+    with {:ok, to_return, buffer, body_remaining} <-
+           do_read_content_length_body(req.socket, req.buffer, body_remaining, opts) do
+      if buffer == <<>> && body_remaining == 0 do
```
asakura, line 204: I'm not so sure about comparing with empty binary as efficient as calling to `byte_size/1`'s BIF

lib/bandit/http1/adapter.ex
```diff
@@ -330,26 +306,24 @@ defmodule Bandit.HTTP1.Adapter do
...
-          non_neg_integer(),
-          timeout()
-        ) :: {:ok, iolist()} | {:error, :closed | :timeout | :inet.posix()}
-  defp read(socket, to_read, already_read, read_size, read_timeout) do
+  @dialyzer {:no_improper_lists, read: 4}
+  @spec read(ThousandIsland.Socket.t(), non_neg_integer(), iolist(), keyword()) ::
+          {:ok, iolist()} | {:error, :closed | :timeout | :inet.posix()}
+  defp read(socket, to_read, already_read, opts) do
+    read_size = Keyword.get(opts, :read_length, 1_000_000)
+    read_timeout = Keyword.get(opts, :read_timeout)
```
asakura, line 320: Searching in a keyword list on every read call is not free. Passing timeouts in arguments (registers of VM) is virtually free.

asakura:
> I started with the same realization you did on https://github.com/mtrudel/bandit/pull/183, and then realized that the body_remaining: 0 function clause could disappear entirely if it collapsed into the primary content-length read_req_body clause. Then I realized that I could de-dupe a bunch of the return handling by moving that out to a trailing with block.

Overall I really liked the initial version (before forced-push to the branch). It was neat and clever refactoring. :)

lib/bandit/http1/adapter.ex
mtrudel: Very good point. I think the tradeoff here is that we only actually access the keywords when we *actually* read off the wire, vs before when we read them every time we called read_req_body. Moreover, we'll almost always be doing a handful of read calls (the default length to read is 1MB at a time). So I *think* this will actually net to fewer keyword calls.

lib/bandit/http1/adapter.ex
mtrudel: Yep, good call.

lib/bandit/http1/adapter.ex
mtrudel: Good call. Two size calls to calculate the difference will probably be cheaper.

test/bandit/http1/request_test.exs
```diff
@@ -483,27 +483,79 @@ defmodule HTTP1RequestTest do
...
+
+    test "handles the case where we ask for more than is already in the buffer", context do
+      client = SimpleHTTP1Client.tcp_client(context)
+
+      :gen_tcp.send(
+        client,
+        "POST /beyond_buffer_read HTTP/1.1\r\nhost: localhost\r\ncontent-length: 5\r\n\r\nAB"
+      )
+
+      Process.sleep(100)
```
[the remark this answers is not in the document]
mtrudel: The intent here is to build up a pretty synthetic situation to force the read logic down specific paths (during the workup of this ticket I'd added explicit `IO.puts` calls to ensure such paths were taken, but we don't have a way to programmatically validate that in tests). Part of that work is ensuring that the server has read a known portion of the request body at each step.

test/bandit/http1/request_test.exs
```diff
@@ -483,27 +483,79 @@ defmodule HTTP1RequestTest do
...
+
+      :gen_tcp.send(
+        client,
+        "POST /in_buffer_read HTTP/1.1\r\nhost: localhost\r\ncontent-length: 5\r\n\r\nABCDE"
+      )
+
+      assert {:ok, "200 OK", _, "A,BCDE"} = SimpleHTTP1Client.recv_reply(client)
+    end
+
+    def in_buffer_read(conn) do
```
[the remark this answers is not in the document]
mtrudel: Yeah, the test module is its own Plug that reads in the request path and calls `apply` against it. Implementation is [here](https://github.com/mtrudel/bandit/blob/main/test/support/server_helpers.ex#L48)

lib/bandit/http1/adapter.ex
asakura: There are few counter arguments I could think of
* file uploads: upload of 100 megs equals to 200 keyword looks up which of course don't put that much overhead, but taking into account that there are only 2 places in module from where `read` is called, 200 calls cloud have been reduces just to 4.
* is :read_length configurable by a user? If so, someone might try reduce its value to smaller size to match for example their network stack configuration parameters (for some users latency is more important that throughput and they literally re-configuring the entire stack  to make their systems react faster in exchange some much they can process)

BTW is that `opts` coming from Plug dominion might change between calls? If no, why can't it be carried along with `THousandIsland.Socket` struct?

test/bandit/http1/request_test.exs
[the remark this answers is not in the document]
asakura: Clear, thanks!

test/bandit/http1/request_test.exs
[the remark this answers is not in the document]
asakura: Clear, thanks!

jfyi I think it's possible to trace with Erlang which functions a process calls. Traces can be simply `assert_received` in ExUnit test. It would be fun to try it out with Elixir once.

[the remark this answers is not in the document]
mtrudel: That *would* be cool! I'm trying to keep tests as covering combinations of externally observable behaviour and not a run through of the internal implementation (ie: the tests are 'what happens if the plug asks for small amounts of data that we already have in buffer', or 'what happens if the plug asks for data that is beyond the scope of the buffer'?) so mandating specific lines of execution isn't *exactly* what we want here.

lib/bandit/http1/adapter.ex
mtrudel: The `opts` here is as passed via [`Plug.Conn.read_body`](https://hexdocs.pm/plug/Plug.Conn.html#read_body/2), so it varies from call to call.

mtrudel: Fair point - I'll refactor them up to the point of first use, and get the best of both worlds.

mtrudel: Updated

lib/bandit/http1/adapter.ex
```diff
@@ -288,8 +247,31 @@ defmodule Bandit.HTTP1.Adapter do
...
+    max_desired_bytes = Keyword.get(opts, :length, 8_000_000)
+
+    if byte_size(buffer) >= max_desired_bytes || body_remaining == 0 do
+      # We can satisfy the read request entirely from our buffer
+      bytes_to_return = min(max_desired_bytes, byte_size(buffer))
+      <<to_return::binary-size(bytes_to_return), rest::binary>> = buffer
+      {:ok, to_return, rest, body_remaining}
+    else
+      # We need to read off the wire
+      bytes_to_read = min(max_desired_bytes - byte_size(buffer), body_remaining)
```
mtrudel, line 261: Special handling needed here for #174 when the time comes.

mtrudel:
Fixing to merge this; @asakura / @moogle19 / @ryanwinchester you have any further thoughts here?

lib/bandit/http1/adapter.ex
```diff
@@ -288,8 +247,31 @@ defmodule Bandit.HTTP1.Adapter do
...
+      read_size = Keyword.get(opts, :read_length, 1_000_000)
+      read_timeout = Keyword.get(opts, :read_timeout)
+
+      with {:ok, iolist} <- read(socket, bytes_to_read, [], read_size, read_timeout) do
+        to_return = IO.iodata_to_binary([buffer | iolist])
+        body_remaining = body_remaining - (byte_size(to_return) - byte_size(buffer))
```
asakura, line 267: Could you run tests please with:
```
        unquote(
          if Mix.env() in [:dev, :test] do
            quote do
              if var!(body_remaining) < 0 do
                raise "body_remaining < 0: #{var!(body_remaining)}"
              end
            end
          end
        )
```

just to be sure that the counter never go below 0

lib/bandit/http1/adapter.ex
asakura: Thank you. I believe that this thread is resolved :)

asakura:
@mtrudel just a minor one above. The PR looks awesome to me!

lib/bandit/http1/adapter.ex
mtrudel: Yep - this is a loose end I'm aware of and was planning on tackle as part of #174. I'll add a failing test that exercises this & we can skip it for now (I'm planning on tackling 174 later this afternoon so it won't be long)

asakura: Ok, no problem then. Let me know please if I can assist you.

Issue #91, opened 2023-02-03
Add HTTP/2 h2c Support, and HTTP/2 Websockets


jeregrine:
I was talking to @mtrudel on slack and he came up with a list of steps in order to support http/2 h2c and http2 websockets:

1. Add support for upgrade support to bandit/http1/{adapter,handler}.ex. This can mostly be cribbed directly from websocket support. I wouldn’t worry about parsing the request details here; just assume that a Plug will parse them out and pass it through as part of the opts in some format.
2. Get it ending up as a {:switch, Bandit.HTTP2.Handler, opts} call that passes the opts through to the HTTP2 handler
3. Add support to bandit/http2/hadnler.ex ’s handle_connection/2 function for starting up an HTTP2 request based on the opts passed in above. Per RFC7540§3.2 this will need to know about settings & the HTTP/1 request that included the upgrade request.
4. Implement a Plug that knows how to read Upgrade requests, package the request & settings header up, and call the Plug.Conn.upgrade_adpater/4 that was added in step 1 above.
5. Finally, we can shim this plug in as part of the overall plug pipeline internal to Bandit, at least for HTTP/1 requests.

May make sense to tackle step 4 first, depending on how you break the problem down in your head

mtrudel:
Happy to take on any / all of this, but @jeregrine indicated they may be taking some on first. 

Let's coordinate here.

mtrudel:
h2c support just landed in 1.0.0-pre.10 thanks to the hard work of @alisinabh! Let us know if you're able to make productive use of it @jeregrine ! 

RFC 8441 support is a separate (and in progress) workup, slated to land as part of 1.1, most likely later this summer. I'll update the scope of this issue to reflect that support; if there are any issues with h2c support please open them as a separate issue.

jeregrine:
WHOA NICE! @alisinabh !!

mtrudel:
> WHOA NICE! @alisinabh !!

One hell of a first-time contribution, amirite?

alisinabh:
I'm flattered guys :) 
To be honest, bandit code is super easy to understand. Thank you all and looking forward to help with other tasks here.

michealp-coder:
Is this still an open issue?

mtrudel:
> Is this still an open issue?

It is! It's a bit of a yak shave to land this properly, which is why it's taking so long. Current progress:

[...]

PR #186, opened 2023-06-25
Support h2c upgrade

alisinabh:
Refs #91 

This adds h2c upgrade support to bandit. Also a small refactoring was done for the `SimpleH2Client` and `SimpleHTTP1Client` to support both transports (TCP and SSL) for both of the protocols. However this can be dropped easily from this PR in case it is too much clutter or unwanted.

Thanks!


mtrudel:
This is outstanding work! Aside from a few *really* small niggles this is ready to go as is. Really well done.

Could you split out the existing test refactoring to a separate PR? It looks like you've already got them broken out by commits. I would suggest 1. opening up a new PR based off of 4755eba2e57652f6a304a1a9d969798f412e53a2, 2. updating this PR to use that one as a base.

The test PR can come across right away I think. In the meantime we can work through the few changes on this one.

lib/bandit/http1/handler.ex
```diff
@@ -54,6 +56,16 @@ defmodule Bandit.HTTP1.Handler do
             )
 
           {:switch, Bandit.WebSocket.Handler, state}
+
+        {:ok, :h2c, remote_settings_payload, method, request_target, headers} ->
```
mtrudel, line 60: I wonder if this signature would be more consistent looking like 
```
 {:ok, :h2c, remote_settings_payload, {method, request_target, headers}} ->
```

WDYT?

lib/bandit/http1/handler.ex
```diff
@@ -88,6 +100,50 @@ defmodule Bandit.HTTP1.Handler do
...
   end
 
+  defp maybe_upgrade_h2c(req, transport_info, method, request_target, headers) do
+    with {:upgrade, "h2c"} <- {:upgrade, Bandit.Headers.get_header(headers, "upgrade")},
+         %Bandit.TransportInfo{secure?: false} <- transport_info,
+         {:ok, connection_headers} <- get_connection_header_keys(headers),
+         {:settings, [{"http2-settings", settings_payload}]} <-
+           {:settings, Enum.filter(headers, fn {key, _value} -> key == "http2-settings" end)},
+         {:ok, settings_payload} <- Base.url_decode64(settings_payload) do
+      Bandit.HTTP1.Adapter.send_resp(
```
mtrudel, line 110: Does this not format onto a single line?

lib/bandit/http1/handler.ex
```diff
@@ -88,6 +100,50 @@ defmodule Bandit.HTTP1.Handler do
...
+      value ->
+        header_keys =
+          value
+          |> String.downcase()
+          |> String.split(",")
+          |> Enum.map(&String.trim/1)
+
+        {:ok, header_keys}
+    end
+  end
```
mtrudel, line 145: This probably belongs in Bandit.Header. I think there may be a helper for this in Plug.Utils as well?

lib/bandit/http1/handler.ex
```diff
@@ -88,6 +100,50 @@ defmodule Bandit.HTTP1.Handler do
...
+      headers =
+        Enum.reject(headers, fn {key, _value} ->
+          key == "connection" || key in connection_headers
+        end)
+
+      {:ok, :h2c, settings_payload, method, request_target, headers}
+    else
+      %Bandit.TransportInfo{secure?: true} -> {:error, "h2c is only supported on http"}
+      {:upgrade, _} -> {:ok, :no_upgrade}
+      {:settings, _} -> {:error, "Expected 1 http2-settings header"}
```
mtrudel, line 126: Can you reference the source RFC in the error message (RFC123§1.2.3)?

lib/bandit/http2/connection.ex
```diff
@@ -51,27 +51,35 @@ defmodule Bandit.HTTP2.Connection do
         }
 
   @spec init(Socket.t(), Bandit.Pipeline.plug_def(), keyword()) :: {:ok, t()} | no_return()
-  def init(socket, plug, opts) do
+  def init(socket, plug, opts, initial_request \\ nil, remote_settings_payload \\ <<>>) do
```
mtrudel, line 54: There's a dialyzer failure in CI on this function

alisinabh:
@mtrudel Thank you for the quick review. I will shortly address the comments.

> 2. updating this PR to use that one as a base.

I have create a new PR (#188) but I cannot base this off of that since the branch only lives in my fork (or maybe I don't know how). Any suggestions how to do it?

lib/bandit/http1/handler.ex
alisinabh: Great. Didn't know about the util function. Done in 5152ee0

(However note that `Plug.Conn.Utils.list/1` in plug version 1.14.2 does not trim trailing spaces but the main branch is now fixed)

lib/bandit/http1/handler.ex
alisinabh: Sure. Here it is 5152ee0

lib/bandit/http2/connection.ex
alisinabh: Sorry forgot about the spec for this function. Fixed in 5152ee0

alisinabh:
Yep, looks like it needed a rebase. Diff looks ok now.

test/bandit/http2/h2c_test.exs
```diff
@@ -0,0 +1,71 @@
...
+      {:ok, _send_ctx} =
+        SimpleH2Client.send_simple_headers(client, 3, :get, "/echo_components", context.port)
+
+      assert {:ok, 3, false, _headers, _recv_ctx} = SimpleH2Client.recv_headers(client, recv_ctx)
+      {:ok, 3, true, body} = SimpleH2Client.recv_body(client)
+
+      assert Jason.decode!(body)["path_info"] == ["echo_components"]
+      assert Jason.decode!(body)["query_string"] == ""
+    end
+  end
```
mtrudel, line 55: We should probably also have a test clause here if the user has disabled HTTP/2 in config

lib/bandit/http1/handler.ex
alisinabh: I have update this (send_resp) in ee4cdfb. Let me know what you think.

alisinabh:
FYI Also found another issue. If the initial request (The one done prior to the upgrade) has body, it will not work! Will be fixing that...

test/bandit/http2/h2c_test.exs
alisinabh: Great catch. Addressed in 02a2fc3b517ceaed5af8caa98b76795f789609b5

lib/bandit/http1/handler.ex
```diff
@@ -88,6 +100,34 @@ defmodule Bandit.HTTP1.Handler do
...
+         {:ok, connection_headers} <- Bandit.Headers.get_connection_header_keys(headers),
+         {:settings, [{"http2-settings", settings_payload}]} <-
+           {:settings, Enum.filter(headers, fn {key, _value} -> key == "http2-settings" end)},
+         {:ok, settings_payload} <- Base.url_decode64(settings_payload),
+         resp_headers = [{"connection", "Upgrade"}, {"upgrade", "h2c"}],
+         {:ok, _sent_body, _req} <- Bandit.HTTP1.Adapter.send_resp(req, 101, resp_headers, <<>>) do
+      headers =
+        Enum.reject(headers, fn {key, _value} ->
+          key == "connection" || key in connection_headers
+        end)
```
mtrudel, line 116: I'm not sure we need to be removing the connection headers (I don't see anything in RFC7540§3.2 about it)?

alisinabh: I did this because of the following specific paragraph from RFC7540§3.2.1

> Since the upgrade is only intended to apply to the immediate
   connection, a client sending the HTTP2-Settings header field MUST
   also send "HTTP2-Settings" as a connection option in the Connection
   header field to prevent it from being forwarded (see Section 6.1 of
   [RFC7230]).

But that is regarding proxy servers and gateways (as per RFC7230§6.1) so probably we can keep them. Just let me know which one it is.

lib/bandit/http1/handler.ex
```diff
@@ -88,6 +100,35 @@ defmodule Bandit.HTTP1.Handler do
...
 
+  defp maybe_upgrade_h2c(state, req, transport_info, method, request_target, headers) do
+    with {:http_2_enabled, true} <- {:http_2_enabled, state.http_2_enabled},
+         {:upgrade, "h2c"} <- {:upgrade, Bandit.Headers.get_header(headers, "upgrade")},
+         %Bandit.TransportInfo{secure?: false} <- transport_info,
+         {:ok, connection_headers} <- Bandit.Headers.get_connection_header_keys(headers),
+         {:settings, [{"http2-settings", settings_payload}]} <-
+           {:settings, Enum.filter(headers, fn {key, _value} -> key == "http2-settings" end)},
+         {:ok, settings_payload} <- Base.url_decode64(settings_payload),
+         {:ok, body, req} <- Bandit.HTTP1.Adapter.read_req_body(req, []),
```
alisinabh, line 111: To handle the initial request, we need to read the request body in HTTP/1.1 (as per RFC7540§3.2)

To do that I am using `Bandit.HTTP1.Adapter.read_req_body/2` and used an empty list for the options. I don't like this but didn't find a better solution. One problem with this is the read_body options will be ignored (e.g. read_timeout and read_length).

lib/bandit/http1/handler.ex
alisinabh: Also worth mentioning: RFC9113§8.2.2

mtrudel: Yeah, we should be keeping them. Whether or not those headers get forwarded (or what gets done with them in any way, really) is the Plug's call, not ours.

alisinabh: But if we don't remove the "connection" header here and pass it through, then [stream_task.ex:151](https://github.com/mtrudel/bandit/blob/main/lib/bandit/http2/stream_task.ex#L151) will complain!

mtrudel: Ah! Good point! I'll have a closer look at this tomorrow morning (~14:00 UTC)

lib/bandit/http1/handler.ex
mtrudel: This is actually more than a bit tricky as we need a bit of logic to handle `{:more,...}` and other tuples. Something like:

```
{:ok, body, req} <- do_read_req_body(req)

.....

# This function is only used during the h2c upgrades
defp do_read_req_body(req, acc \\ <<>>) do
  case Bandit.HTTP1.Adapter.read_req_body(req, []) do
    {:ok, body, req} -> {:ok, acc <> body, req}
    {:more, body, req} -> do_read_req_body(req, acc <> body)
    {:error, error} -> {:error, error}
  end
end 
```

lib/bandit/http1/handler.ex
mtrudel: Yep, your original implementation is correct; we should be stripping those.

lib/bandit/http2/connection.ex
```diff
@@ -513,4 +529,40 @@ defmodule Bandit.HTTP2.Connection do
...
+             connection.telemetry_span,
+             headers,
+             true,
+             connection.plug,
+             body,
+             connection.opts
+           ),
+         stream = %{stream | state: :remote_closed},
+         {:ok, streams} <- StreamCollection.put_stream(connection.streams, stream) do
+      {:ok, %{connection | streams: streams}}
```
mtrudel, line 553: I wonder if we shouldn't be DRYing this up with the existing header receiving logic in the `handle_frame(%Header{})` function above (basically have `handle_frame(%Header{})` handle the header decompression only, and pull most of that function out to an internal `handle_headers()` function that we could reuse here).

lib/bandit/http2/connection.ex
```diff
@@ -513,4 +529,40 @@ defmodule Bandit.HTTP2.Connection do
...
+         true <- accept_headers?(headers, connection.opts, stream),
+         {:ok, stream} <-
+           Stream.recv_headers(
+             stream,
+             connection.transport_info,
+             connection.telemetry_span,
+             headers,
+             true,
+             connection.plug,
+             body,
```
mtrudel, line 548: I'm pretty sure we don't need all the `premature_body` plumbing here - you should be able to just doctor up a data frame with the body and call `handle_frame(%Data{})`. There's a fair bit of window space accounting in there. 

Mind you, the RFC doesn't mention anything about whether or not the initial body should be counted against any of the windows. Regardless, I think it's possible for us to do that and remove a whole bunch of the diffs here against stream & adapter (I don't think either of those files should need to change at all).

lib/bandit/http2/connection.ex
alisinabh: Addressed in b8cf932

lib/bandit/http2/connection.ex
alisinabh: Yep you are right there were no change required for adapter and stream. I have addressed this in 23b526220ee8a44d9e46d6b3b58df2d3785b2408

lib/bandit/http1/handler.ex
ryanwinchester: > (However note that `Plug.Conn.Utils.list/1` in plug version 1.14.2 does not trim trailing spaces but the main branch is now fixed)

There is a skipped test related to this as well. 

See:
https://github.com/mtrudel/bandit/blob/main/test/bandit/headers_test.exs#L69-L78

lib/bandit/http2/connection.ex
```diff
@@ -513,4 +532,29 @@ defmodule Bandit.HTTP2.Connection do
...
+    {_, _, _, path} = request_target
+    headers = [{":scheme", "http"}, {":method", method}, {":path", path} | headers]
+
+    with {:ok, stream} <- StreamCollection.get_stream(connection.streams, 1),
+         {:ok, stream} <- handle_headers(headers, stream, true, connection),
+         {:ok, stream, _stream_window_increment} <- Stream.recv_data(stream, data),
+         {:ok, stream} <- Stream.recv_end_of_stream(stream, true),
+         {:ok, streams} <- StreamCollection.put_stream(connection.streams, stream) do
+      {:ok, %{connection | streams: streams}}
+    else
```
mtrudel, line 546: This is real nice. I'm glad we kept pushing on improving this!

mtrudel:
I added a couple of test changes (and one small fix to body size checks). LMK when you've had a chance to review @alisinabh and we should be good to merge this!

vanderhoop:
@mtrudel @alisinabh Given that h2c is deprecated, would it make sense to throw the `benchmark` label on here to make sure that the perf cost (if any) is known before merge?

alisinabh (commented):
Thanks @mtrudel, Everything looks good.

One last question: Do we need a switch to disable h2c?

mtrudel:
Heads up that I'm undertaking a pretty major refactor of HTTP/2 at #286, and there's a small change to our stream receive window handling logic; the changes I'm making make it so that we now DO count the initial body against the stream's receive window. Recall that RFC 7540 wasn't explicit about this, but since they go to pains to mention that the initial request body is received over HTTP/1, I think it's implied that it's not *supposed* to count against the receive window. However, this whole thing is a bit moot, since:

1. The spec is clear that the stream is to be treated as remote closed initially (ie: the client is unable to send more data on the stream). So there's no chance of us worrying about the remote end sending more data.
2. We can send whatever window resets we want within the spec; we're not beholden to NOT send them in this situation, so the fact that we're updating the receive window on a stream that's already closed in the receive direction is at most inefficient, but still correct.

PR #193, opened 2023-07-07
Fix Typespecs

moogle19:
Fix some typespecs and add cases for `:undefined` and `:unspec` in `bound_address/2`

### DIFF
--- a/lib/bandit/pipeline.ex
+++ b/lib/bandit/pipeline.ex
@@ -90,7 +90,7 @@ defmodule Bandit.Pipeline do
        ),
        do: {:ok, to_string(host), port || determine_local_port(local_info)}
 
-  @spec determine_local_port(ThousandIsland.Transport.socket_info()) :: integer()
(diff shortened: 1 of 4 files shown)


lib/bandit/pipeline.ex
```diff
@@ -90,7 +90,7 @@ defmodule Bandit.Pipeline do
        ),
        do: {:ok, to_string(host), port || determine_local_port(local_info)}
 
-  @spec determine_local_port(ThousandIsland.Transport.socket_info()) :: integer()
+  @spec determine_local_port(ThousandIsland.Transport.socket_info()) :: :inet.port_number()
```
ryanwinchester, line 93: Since Bandit is designed for `Plug` should it be `Plug.Conn.port_number()`?
I know it's the same thing in the end, so it is of little importance, but thought I'd ask anyway.

moogle19: Good point. Just updated the PR.
Thanks!

mtrudel:
Rad!

PR #194, opened 2023-07-07
Fix a typo in README.md

dorian-marchal:


mtrudel:
Thanks for the catch!

PR #198, opened 2023-07-14
Improved error message when start_link is called without :plug option.

JBraungardt:


mtrudel:
Thanks! Typos are teh worst.

PR #207, opened 2023-08-07
Improve RFC compliance regarding cacheability on deflated responses

### LINKED ISSUE #203: Compress in compliance with RFCs
Great work with this project!

Working on [HTTP caching](https://github.com/tanguilp/plug_http_cache), I came upon some potential issues with HTTP compression in Cowboy that seem to apply to the current project as well:
- when compression is enabled, the `content-encoding` header should be added to the vary header (https://github.com/ninenines/cowboy/issues/1529). Actually it seems `vary` is never set. Since both `gzip` and `deflate` are supported (in addition to no encoding), it seems necessary so as to not confuse caches
- when the uncompressed response has a strong etag, the compressed response should have a different etag (https://github.com/ninenines/cowboy/issues/1546)
- I'm not sure I'm seeing `content-length` being reset after compression
- the [`no-transform` cache control directive](https://www.rfc-editor.org/rfc/rfc9111.html#name-no-transform-2) might apply to bandit

mtrudel:
Fixes #203:

* Include a `vary: accept-encoding` header when deflating responses
* Do not deflate responses which contain a strong etag
* Do not deflate responses which contain a `no-transform` directive in the `cache-control` header


mtrudel:
@tanguilp relevant to your interests

moogle19:
Shouldn't `vary: accept-encoding` also be set when the response isn't compressed, but the server could send a compressed response?
Wouldn't a CDN (or what ever caches the request) otherwise always serve the uncompressed version, when it was requested first?

tanguilp:
> Shouldn't `vary: accept-encoding` also be set when the response isn't compressed, but the server could send a compressed response?

That's correct. As soon as compression is enabled, response should vary on `content-encoding` even for uncompressed responses (see https://github.com/ninenines/cowboy/issues/1529 for further resources).

mtrudel:
Ah yes, good point! I'll get that fixed in the PR

tanguilp:
I've had a quick glance and it looks good to me :shipit:

lib/bandit/http1/adapter.ex
```diff
@@ -370,6 +385,8 @@ defmodule Bandit.HTTP1.Adapter do
           {body, headers, %{}}
       end
 
+    compress = Keyword.get(req.opts.http_1, :compress, true)
```
moogle19, line 388: Could we avoid a keyword lookup here? 
If I remember correctly, this was quite slow and would be executed for every request.

mtrudel: I'm not actually going to address this here since I'm planning on doing a broader pass over configuration to be consistent with this everywhere.

PR #211, opened 2023-08-14
Add support for sending messages on websocket close

mtrudel:
Implements https://github.com/phoenixframework/websock/pull/12, pending a release of websock and a dependency bump here


lib/bandit/websocket/connection.ex
```diff
@@ -205,6 +205,10 @@ defmodule Bandit.WebSocket.Connection do
       {:stop, :normal, code, websock_state} ->
         do_stop(code, :normal, socket, %{connection | websock_state: websock_state})
 
+      {:stop, :normal, code, msg, websock_state} ->
+        _ = do_deflate(msg, socket, %{connection | websock_state: websock_state})
```
icehaunter, line 209: I have two questions about ignoring the return value here:
1. If `do_deflate` succeeds, then we're ignoring updated metrics and `stop_span/2` call within `do_stop/4` will report slightly incorrect results
2. If `do_deflate` fails, we're ignoring a continuation with updated `state: :closed`, so `do_stop/4` will call WebSock `terminate/2` callback a second time on the same connection, and then try to close the socket and span for a second time (not sure if this will cause any errors, haven't dug into `Socket.close/1` code

Mildly incorrect metrics are not the end of the world, but calling `terminate/2` twice may lead to some unexpected behaviour I think

icehaunter:
Awesome functionality, just what we were missing!

lib/bandit/websocket/connection.ex
mtrudel: Good point about the metrics - I was thinking mostly of failure cases (and, if I'm being honest, just being a bit lazy). 

I'll get those fixed up.

mtrudel: Fixed.

PR #216, opened 2023-08-23
Return errors from chunk/2 calls in HTTP/1

### LINKED ISSUE #214: chunk calls ignore send errors
https://github.com/mtrudel/bandit/blob/b27d4416fe2aed37b4baff2cd95fe680f56a8427/lib/bandit/http1/adapter.ex#L468C34-L469

It looks like, at least for the http1 adapter (and maybe for the http2 adapter also), `chunk/2` calls ignore errors from the socket layer. I found this testing some streaming code which I was testing with `curl`. The streaming itself all worked fine, but if terminated `curl` there was no way for my plug to discover this since `chunk/2` _always_ returns `:ok`. Since, in this case, the intention is to keep streaming until the client terminates, this leaves upstream processes running for some period of time (possibly indefinitely)

mtrudel:
Fixed #214


mtrudel:
Relevant to your interests @ausimian

PR #221, opened 2023-09-03
Update Req

wojtekmach:
Hey, thanks for using Req for integration testing Bandit!

The main cause of the churn was changing headers to be maps: https://github.com/wojtekmach/req/releases/tag/v0.4.0. Note, they can still be set in `Req.new(headers: ...)`, `Req.get!(..., headers: ...)` as lists of tuples, however they are internally normalised and stored as maps of string names to lists of values.

In the ReqHelpers, instead of disabling _some_ things I think it is cleaner (if verbose) to start with a blank slate and only enable what you're gonna need.


mtrudel:
Thanks for the thoughtful work, Wojtek! Currently camping, but I’ll review / merge this as soon as I’m back in civilization.

test/bandit/http1/request_test.exs
```diff
@@ -351,7 +351,7 @@ defmodule HTTP1RequestTest do
       response =
         Req.get!(context.req,
           url: "/expect_headers/a//b/c?abc=def",
-          headers: [{"X-Fruit", "banana"}]
+          headers: [{"x-fruit", "banana"}]
```
wojtekmach, line 354: Req automatically downcases header names (and so does Mint. which is doing the same work twice and yeah I'd like to do something about it.) so I thought being more explicit here would be helpful.

PR #228, opened 2023-09-12
Use protocol default port in the event that no port is provided in host

### LINKED ISSUE #106: Host port should default to 80 if the Host header has been set
Assuming we want to keep compatibility with cowboy:

When the host header is set with cowboy without an explicit port, cowboy will default to port 80. 

To reproduce:

```bash
curl -H "Host: foo.co.za" http://localhost:4000
```
Should set conn.port to 80

```bash
curl http://localhost:4000
```
Should set conn.port to 4000

and 

```bash
curl -H "Host: foo.co.za:9999" http://localhost:4000
```
Should set conn.port to 9999

This is currently a issue with ueberauth when generating the callback_url.

mtrudel:
Fixes #106

Relevant to your interests @wkirschbaum

### DIFF
--- a/lib/bandit/pipeline.ex
+++ b/lib/bandit/pipeline.ex
@@ -35,8 +35,7 @@ defmodule Bandit.Pipeline do
   defp build_conn({mod, req}, transport_info, method, request_target, headers) do
     with {:ok, scheme} <- determine_scheme(transport_info, request_target),
          version <- mod.get_http_protocol(req),
-         {:ok, host, port} <-
-           determine_host_and_port(transport_info, version, request_target, headers),
+         {:ok, host, port} <- determine_host_and_port(scheme, version, request_target, headers),
          {path, query} <- determine_path_and_query(request_target) do
       uri = %URI{scheme: scheme, host: host, port: port, path: path, query: query}
       %Bandit.TransportInfo{peername: {remote_ip, _port}} = transport_info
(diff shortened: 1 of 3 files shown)


wkirschbaum:
Thank you, this change means that the host port won't be converted to the underlying transport port. I feel we are still missing a piece when it comes to load balancers and identifying x-forward-for headers, but outside the scope of this discussion and probably this library. 

I will test this on my side in the next couple of days.

yordis:
> it would be helpful if you could rephrase this as a concrete spec
question and then send it to the HTTP WG's mailing list
(https://lists.w3.org/Archives/Public/ietf-http-wg/).
> FWIW, <https://greenbytes.de/tech/webdav/rfc9112.html#absolute-form>

I got some help from an editor of RFC 9110

mtrudel:
@yordis we have pretty comprehensive coverage of absolute URIs (https://github.com/mtrudel/bandit/blob/de175959d5f62293b9e923f85847e1cf0e24e2dd/test/bandit/http1/request_test.exs#L195-L254 for HTTP/1, https://github.com/mtrudel/bandit/blob/de175959d5f62293b9e923f85847e1cf0e24e2dd/test/bandit/http2/protocol_test.exs#L2131-L2237 for H2). 

I've re-read the relevant spec links above a few times over the last couple of days and this is sitting better and better with me. Quite confident we're doing it right here.

Issue #230, opened 2023-09-18
Unable to turn on compression on http1 adapter


travelmassive:
I can't get gzip compression working on Gigalixir (1.0.0-pre.15), so I did some quick debugging.

Gigalixir handles SSL traffic, so I looked at the http1 adapter (assuming it reverse proxies to http) and tested on my local dev.

I set ```http_1_options: [compress: true]``` in my config as per the docs, and didn't get a compressed response. 

It looks like there might be a logical bug here:

lib/bandit/http1/adapter.ex, line 376 in send_resp:

```
when byte_size(body) > 0 and not is_nil(content_encoding) ->
```

changing to (copying from http2/adapter.ex) works:

```
when body != <<>> and not is_nil(content_encoding) ->
```

Thanks for any assistance! - Ian

mtrudel:
I'm not immediately sure how that change would affect anything; for the purposes of this use those amount to the same thing (comparisons to the empty binary are generally to be avoided for performance reasons; I'll review why we're doing that in http2). 

There are a number of places where we purposely do not cache responses, even if configured. I'd suggest reviewing the clauses in #207 and #203 to see if you're possibly hitting one of those cases.

Semi-relatedly: you DID however make me notice that the `byte_size` should in fact be `IO.iodata_length` since it's entirely valid for the user to pass an iolist in as `body`.  I'll get that fixed directly.

travelmassive:
Examining the cases of not compressing was how I stumbled on this — I was examining the logic inside send_resp and got an error ("1st argument: not a bitstring") trying ```IO.inspect(byte_size(body))```. 

For my local environment I'm connecting direct via http://dev.local:4000 with Chrome, and there's no etag header or "no-transform" directive in the cache-control header in my request. So I'd expect that compression should be enabled.

mtrudel:
I think what you were / are seeing was the same thing I came up against testing this, namely that calling `byte_size(a) > 0` in a guard will silently fail since [errors are failures](https://hexdocs.pm/elixir/main/patterns-and-guards.html#errors-in-guards) within guards (TBH, I don't know that I ever actually *knew* this before today).  

Based on the fact that you're getting that error trying to inspect your body, I'd bet you're (legally!) trying to send an iolist body, which is failing the guard on the compression arm of that case statement. Neat!

I'm merging #231 now and will be cutting `-pre.16` in a few minutes.

Thanks for a great bug!

mtrudel:
[...]

PR #231, opened 2023-09-18
Improve handling of iolist response bodies

mtrudel:
As noted in #230, there were a few places where we weren't optimally deflating iolist response bodies. This PR fixes those cases, as well as updates a few comparisons of the form `a == <<>>` with `byte_size(a) == 0` where we know for sure that `a` is a binary.


travelmassive:
Can confirm this PR fixes the issue I reported with not being able to enable compression for http1 adapter.

PR #236, opened 2023-09-19
Support function based plugs

### LINKED ISSUE #234: Add support for function plugs
The Plug specification distinguishes between two type of plugs: function and module plugs. See https://hexdocs.pm/plug/Plug.html

It seems Bandit only supports module plugs, and passing a function to the `plug` argument fails with:

```
     ** (RuntimeError) failed to start child with the spec {Bandit, [plug: #Function<1.50843548/2 in Bernard.Webserver.start!/1>, scheme: :http, port: 10000]}.
     Reason: an exception was raised:
         ** (ArgumentError) errors were found at the given arguments:
     
       * 1st argument: not an atom
     
             :erlang.apply(#Function<1.50843548/2 in Bernard.Webserver.start!/1>, :init, [[]])
             (bandit 1.0.0-pre.15) lib/bandit.ex:307: Bandit.plug/1
             (bandit 1.0.0-pre.15) lib/bandit.ex:222: Bandit.start_link/1
             (stdlib 5.0.1) supervisor.erl:420: :supervisor.do_start_child_i/3
             (stdlib 5.0.1) supervisor.erl:406: :supervisor.do_start_child/2
             (stdlib 5.0.1) supervisor.erl:712: :supervisor.handle_start_child/2
             (stdlib 5.0.1) supervisor.erl:461: :supervisor.handle_call/3
             (stdlib 5.0.1) gen_server.erl:1113: :gen_server.try_handle_call/4
             (stdlib 5.0.1) gen_server.erl:1142: :gen_server.handle_msg/6
             (stdlib 5.0.1) proc_lib.erl:251: :proc_lib.wake_up/3

```

mtrudel:
Fixes #234

### DIFF
--- a/lib/bandit.ex
+++ b/lib/bandit.ex
@@ -303,9 +303,19 @@ defmodule Bandit do
     |> Keyword.get(:plug)
     |> case do
       nil -> raise "A value is required for :plug"
-      {plug, plug_options} -> {plug, plug.init(plug_options)}
-      plug -> {plug, plug.init([])}
+      {plug_fn, plug_options} when is_function(plug_fn, 2) -> {plug_fn, plug_options}
+      plug_fn when is_function(plug_fn) -> {plug_fn, []}
+      {plug, plug_options} when is_atom(plug) -> validate_plug(plug, plug_options)
+      plug when is_atom(plug) -> validate_plug(plug, [])
+      other -> raise "Invalid value for plug: #{inspect(other)}"
     end
+  end
+
+  defp validate_plug(plug, plug_options) do
+    if !function_exported?(plug, :init, 1), do: raise("plug module does not define init/1")
+    if !function_exported?(plug, :call, 2), do: raise("plug module does not define call/2")
+
+    {plug, plug.init(plug_options)}
   end
 
   @spec parse_as_number(binary() | integer()) :: integer()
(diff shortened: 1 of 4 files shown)


lib/bandit.ex
```diff
@@ -303,6 +303,8 @@ defmodule Bandit do
     |> Keyword.get(:plug)
     |> case do
       nil -> raise "A value is required for :plug"
+      {plug_fn, plug_options} when is_function(plug_fn) -> {plug_fn, plug_options}
+      plug_fn when is_function(plug_fn) -> {plug_fn, []}
       {plug, plug_options} -> {plug, plug.init(plug_options)}
       plug -> {plug, plug.init([])}
```
moogle19, line 309: Should we us `is_function(plug_fn, 2)` to ensure the function has the correct arity?
Maybe checking if the plug is an atom here would be good as well (or even if the Module exists / is loaded), like:

```elixir
case do
  nil ->
    raise "A value is required for :plug"

  {plug_fn, plug_options} when is_function(plug_fn, 2) ->
    {plug_fn, plug_options}

  plug_fn when is_function(plug_fn, 2) ->
    {plug_fn, []}

  {plug, plug_options} when is_atom(plug) ->
    unless function_exported?(plug, :call, 2) do
      raise "Invalid plug module: #{inspect(plug)}"
    end

    {plug, plug.init(plug_options)}

  plug when is_atom(plug) ->
    unless function_exported?(plug, :call, 2) do
      raise "Invalid plug module: #{inspect(plug)}"
    end

    {plug, plug.init([])}

  _other ->
    raise "Invalid value for :plug"
end
```

This would error on starting the server vs. error on every request, if the function or plug are (obviously) invalid.

mtrudel: Yes we should! Lemme get this in.

mtrudel: Updated!

PR #242, opened 2023-10-04
Fix handling of HEAD responses

### LINKED ISSUE #235: HTTP conformance bug: sends response body on HEAD request
The HTTP specification states that a HEAD request should be functionally identical to a GET one, but shall not include any response body.

I have a simple Plug handler that deals with HEAD and GET requests separately (for reasons). 

```
Webserver
|> on(:head, "/", fn conn ->
  send(parent, :head_received)

  conn
  |> put_resp_content_type("text/html")
  |> resp(200, "OK")
end)
|> on(:get, "/", fn conn ->
  send(parent, :get_received)

  conn
  |> put_resp_content_type("text/html")
  |> resp(200, "OK")
end)
```

The Plug.Conn API doesn't handle the case of sending a response with nil body, so it expects the actual Plug.Conn.Adapter implementation not to send any response body, even if set, if the request method is HEAD. See https://github.com/elixir-plug/plug/blob/2ef07cdd2732cde5cac73fc39b49fe83d5fcc369/lib/plug/conn/adapter.ex#L65

Bandit seems to ignore this suggestion, so will send a response body on HEAD, which might crash any conforming client. In my case I'm using Mint to connect to Bandit, and it fails with `%Mint.HTTPError{reason: :invalid_status_line, module: Mint.HTTP1}` error.

Using the exact same plug with Cowboy works as intended.

mtrudel:
Also fixes a couple of longstanding bugs with 204 & 304 responses being sent with content-length headers & bodies

Fixes #235


lib/bandit/http1/adapter.ex
```diff
@@ -28,6 +29,7 @@ defmodule Bandit.HTTP1.Adapter do
           body_remaining: nil | integer(),
           body_encoding: nil | binary(),
           version: nil | :"HTTP/1.1" | :"HTTP/1.0",
+          method: Plug.Conn.method(),
```
ryanwinchester, line 32: Since `Plug.Conn.method()` is `binary()`?

```suggestion
          method: nil | Plug.Conn.method(),
```

lib/bandit/http2/adapter.ex
```diff
@@ -147,7 +149,7 @@ defmodule Bandit.HTTP2.Adapter do
     headers = Bandit.Headers.add_content_length(headers, body_bytes, status)
 
     adapter =
-      if body_bytes == 0 do
+      if body_bytes == 0 || !send_resp_body?(adapter, status) do
```
ryanwinchester, line 152: They're both booleans

```suggestion
      if body_bytes == 0 or !send_resp_body?(adapter, status) do
```

lib/bandit/http1/adapter.ex
mtrudel: Yup! I did this for http/2 but forgot http/1

lib/bandit/http2/adapter.ex
mtrudel: Lots of prior art for || usage in this case; I generally try to avoid using `or` except when I *have* to

lib/bandit/http2/adapter.ex
```diff
@@ -168,46 +170,45 @@ defmodule Bandit.HTTP2.Adapter do
...
 
-    cond do
-      offset + length == size && offset == 0 ->
-        adapter = send_headers(adapter, status, headers, false)
+    adapter =
+      cond do
+        !send_resp_body?(adapter, status) ->
+          send_headers(adapter, status, headers, true)
+
+        offset + length == size && offset == 0 ->
```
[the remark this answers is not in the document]
mtrudel: ibid.

lib/bandit/http2/adapter.ex
```diff
@@ -168,46 +170,45 @@ defmodule Bandit.HTTP2.Adapter do
...
-      offset + length < size ->
-        with {:ok, fd} <- :file.open(path, [:raw, :binary]),
-             {:ok, data} <- :file.pread(fd, offset, length) do
-          adapter =
+        offset + length < size ->
+          with {:ok, fd} <- :file.open(path, [:raw, :binary]),
+               {:ok, data} <- :file.pread(fd, offset, length) do
             adapter
             |> send_headers(status, headers, false)
             |> send_data(data, true)
```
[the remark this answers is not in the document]
mtrudel: Nice catch! I added similar handling to https://github.com/mtrudel/thousand_island/pull/78/files

lib/bandit/http1/adapter.ex
mtrudel: Done

PR #255, opened 2023-10-24
Add support to :max_heap_size in websocket handler

v0idpwn:
See https://github.com/phoenixframework/websock_adapter/pull/15 for the reasoning. 

Felt like this was necessary here as well, but maybe bandit/thousand_island architecture wouldn't require it, I didn't really look deeply into it.


mtrudel:
Thanks! This looks great!

PR #259, opened 2023-11-02
Tweaks to how Bandit handles unexpected messages

mtrudel:
Previously, unexpected messages received by an HTTP/1 handler would crash the server; this PR updates this behaviour to log messages in the same way that a stock GenServer which does not implement any `handle_info/2` callbacks does, and thence does not crash. 

Note there's no need to add a similar mechanism on HTTP/2 handlers, since HTTP/2  requests are backed (at least for now) by `Task`

Also note that there's no need to add a similar mechanism to WebSockets, since WebSock already provides a way to forward handle_info calls to the websock implementor. In fact, this PR also stops explicitly catching normal monitored process termination calls in the websocket handler for this reason.

### DIFF
--- a/lib/bandit/http1/handler.ex
+++ b/lib/bandit/http1/handler.ex
@@ -197,4 +197,31 @@ defmodule Bandit.HTTP1.Handler do
 
   def handle_info({:EXIT, _pid, :normal}, {socket, state}),
     do: {:noreply, {socket, state}, socket.read_timeout}
+
+  def handle_info(msg, state) do
+    # Copied verbatim from lib/elixir/lib/gen_server.ex
+    proc =
+      case Process.info(self(), :registered_name) do
+        {_, []} -> self()
+        {_, name} -> name
+      end
+
+    :logger.error(
+      %{
+        label: {GenServer, :no_handle_info},
+        report: %{
+          module: __MODULE__,
+          message: msg,
+          name: proc
+        }
+      },
+      %{
+        domain: [:otp, :elixir],
+        error_logger: %{tag: :error_msg},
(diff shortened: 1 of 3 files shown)


lib/bandit/http1/handler.ex
```diff
@@ -197,4 +197,31 @@ defmodule Bandit.HTTP1.Handler do
 
   def handle_info({:EXIT, _pid, :normal}, {socket, state}),
     do: {:noreply, {socket, state}, socket.read_timeout}
+
+  def handle_info(msg, state) do
+    # Copied verbatim from lib/elixir/lib/gen_server.ex
```
mtrudel, line 202: Ideally we'd just call `super(msg, state)` here but calling super on (most) GenServer callbacks was deprecated a [while ago](https://github.com/elixir-lang/elixir/commit/12a4ea4c9fb83dd9f370fc9247207c02cf0d5443#diff-07e9c029fa058bfd89482ef285ba58c16a3f3a64158a3267d8dea0391710da26)

lib/bandit/http1/handler.ex
```diff
@@ -197,4 +197,31 @@ defmodule Bandit.HTTP1.Handler do
...
+    proc =
+      case Process.info(self(), :registered_name) do
+        {_, []} -> self()
+        {_, name} -> name
+      end
```
josevalim, line 207: Do you allow the process to be registered? If not you can skip this, otherwise it looks good to me :)

mtrudel: Yep, handler processes themselves are able to accept any valid gen_server option via the `genserver_options` flag.

PR #268, opened 2023-11-22
Fix Bandit.HTTP1.Adapter.inform/3 return value

wojtekmach:
Per https://hexdocs.pm/plug/Plug.Conn.Adapter.html#c:inform/3, the callback should return:

    :ok | {:error, :not_supported}

The way Plug.Conn.inform! is implemented at the moment, https://github.com/elixir-plug/plug/blob/v1.15.2/lib/plug/conn.ex#L1335, it looked like Bandit did _not_ implement inform but it sure does!

### DIFF
--- a/lib/bandit/http1/adapter.ex
+++ b/lib/bandit/http1/adapter.ex
@@ -505,19 +505,10 @@ defmodule Bandit.HTTP1.Adapter do
   @impl Plug.Conn.Adapter
   def inform(%__MODULE__{version: :"HTTP/1.0"}, _status, _headers), do: {:error, :not_supported}
 
-  def inform(%__MODULE__{socket: socket, version: version} = req, status, headers) do
-    start_time = Bandit.Telemetry.monotonic_time()
-
-    {header_iodata, header_metrics} = response_header(version, status, headers)
+  def inform(%__MODULE__{socket: socket, version: version}, status, headers) do
+    {header_iodata, _header_metrics} = response_header(version, status, headers)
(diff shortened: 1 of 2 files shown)


mtrudel:
Looks good! I'm going to add a small note to the bandit telemetry docs now that we're losing tracking of inform responses but otherwise lgtm!

wojtekmach:
would it help if the adapter allowed you to send `{:ok, payload}` or whatever makes the most sense? (I believe payload is a name used for arbitrary data, at least in adapter.send_resp/4)

If so that would be a compatible change in Plug and I can help pursuing it.

mtrudel:
(Sorry for terse response, mobile). Ideally, yeah. Plug's chunk api also expects a bare ok (no tuple) and it was the cause of a recent oddball bug as well. Ideally everything in adapter would return the same ok tuple shape

wojtekmach:
We need to wait for https://github.com/elixir-plug/plug/pull/1193 in case there's any change in the contract but this is ready. And when there's a new Plug version I'll send a PR that uses the stricter inform!, https://github.com/mtrudel/bandit/pull/268/commits/18e12ed5eaf1e286b0ef5ab561bc8b847a89a17b#diff-6a1dfdf6de8e0ba81a6ae4eef3f45cf79b7d4a55d7130e228463f9f77a0344e3R1339.

mtrudel:
New plug version is here!

mtrudel:
Merged this since it's erroring out on a couple of other PRs without this

PR #272, opened 2023-12-07
Switch to use Erlang nifs in frame mask application.

crertel:
I'd like to throw this strawman up to see if we can maybe switch the XOR work to be done using built-in Erlang nifs.

I _think_ this should result in fewer allocations and should use native code as much as possible for the bulk XORing.

Let me know what y'all think. :)

### DIFF
--- a/lib/bandit/websocket/frame.ex
+++ b/lib/bandit/websocket/frame.ex
@@ -184,28 +184,26 @@ defmodule Bandit.WebSocket.Frame do
     |> IO.iodata_to_binary()
   end
 
-  def mask(payload, mask) do
-    payload
-    |> do_mask(<<mask::32>>, [])
-    |> Enum.reverse()
-    |> IO.iodata_to_binary()
-  end
+  # note that we assume the mask integer is 32-bits and do not check
+  @spec mask( binary(), integer() ) :: binary()
+  def mask(payload, mask_integer) do
 
-  defp do_mask(
-         <<h::unquote(@mask_size), rest::binary>>,
-         <<int_mask::unquote(@mask_size)>> = mask,
-         acc
-       ) do
-    do_mask(rest, mask, [<<Bitwise.bxor(h, int_mask)::unquote(@mask_size)>> | acc])
+    # 1. Allocate the binary for the mask repetitions
+    payload_size = byte_size(payload)
+                
+    mask_repetitions = case { div(payload_size,4), rem(payload_size,4)} do
+      {count_4_bytes, 0 } ->
+        count_4_bytes # payload length is a multiple of 4, so the mask will be fine
+      {count_4_bytes, _count_stragglers} ->
+       count_4_bytes + 1 # bump up by one mask size
+    end
+        
+    mask_binary = :binary.copy( <<mask_integer::32>>, mask_repetitions)
+              
+    # 2. Trim the binary if needed
+    fit_mask = :binary.part(mask_binary, 0, payload_size)
+        
+    # 3. XOR (in a nif) the payload and mask binary
+    masked_payload = :crypto.exor(payload, fit_mask)
   end
-
-  defp do_mask(<<h::32, rest::binary>>, <<int_mask::32, _mask_rest::binary>> = mask, acc) do
-    do_mask(rest, mask, [<<Bitwise.bxor(h, int_mask)::32>> | acc])
-  end
-
-  defp do_mask(<<h::8, rest::binary>>, <<current::8, mask::24, _mask_rest::binary>>, acc) do
-    do_mask(rest, <<mask::24, current::8>>, [<<Bitwise.bxor(h, current)::8>> | acc])
-  end
-
-  defp do_mask(<<>>, _mask, acc), do: acc
 end


lib/bandit/websocket/frame.ex
```diff
@@ -184,28 +184,26 @@ defmodule Bandit.WebSocket.Frame do
...
+    payload_size = byte_size(payload)
+                
+    mask_repetitions = case { div(payload_size,4), rem(payload_size,4)} do
+      {count_4_bytes, 0 } ->
+        count_4_bytes # payload length is a multiple of 4, so the mask will be fine
+      {count_4_bytes, _count_stragglers} ->
+       count_4_bytes + 1 # bump up by one mask size
+    end
+        
+    mask_binary = :binary.copy( <<mask_integer::32>>, mask_repetitions)
```
crertel, line 201: This is where we convert the mask integer out to a bitsring and then duplicate it. I believe this is a NIF and will trigger one allocation.

lib/bandit/websocket/frame.ex
```diff
@@ -184,28 +184,26 @@ defmodule Bandit.WebSocket.Frame do
...
+
+    # 1. Allocate the binary for the mask repetitions
+    payload_size = byte_size(payload)
+                
+    mask_repetitions = case { div(payload_size,4), rem(payload_size,4)} do
+      {count_4_bytes, 0 } ->
+        count_4_bytes # payload length is a multiple of 4, so the mask will be fine
+      {count_4_bytes, _count_stragglers} ->
+       count_4_bytes + 1 # bump up by one mask size
+    end
```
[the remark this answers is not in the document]
crertel: There's probably an even terser way of writing this, but it wasn't coming to me.

ryanwinchester:
```elixir
defmodule Original do
  # Masking is done @mask_size bits at a time until there is less than that number of bits left.
  # We then go 32 bits at a time until there is less than 32 bits left. We then go 8 bits at
  # a time. This yields some significant perforamnce gains for only marginally more complexity
  @mask_size 512

  # Note that masking is an involution, so we don't need a separate unmask function
  def mask(payload, mask) when bit_size(payload) >= @mask_size do
    payload
    |> do_mask(String.duplicate(<<mask::32>>, div(@mask_size, 32)), [])
    |> Enum.reverse()
    |> IO.iodata_to_binary()
  end

  def mask(payload, mask) do
    payload
    |> do_mask(<<mask::32>>, [])
    |> Enum.reverse()
    |> IO.iodata_to_binary()
  end

  defp do_mask(
         <<h::unquote(@mask_size), rest::binary>>,
         <<int_mask::unquote(@mask_size)>> = mask,
         acc
       ) do
    do_mask(rest, mask, [<<Bitwise.bxor(h, int_mask)::unquote(@mask_size)>> | acc])
  end

  defp do_mask(<<h::32, rest::binary>>, <<int_mask::32, _mask_rest::binary>> = mask, acc) do
    do_mask(rest, mask, [<<Bitwise.bxor(h, int_mask)::32>> | acc])
  end

  defp do_mask(<<h::8, rest::binary>>, <<current::8, mask::24, _mask_rest::binary>>, acc) do
    do_mask(rest, <<mask::24, current::8>>, [<<Bitwise.bxor(h, current)::8>> | acc])
  end

  defp do_mask(<<>>, _mask, acc), do: acc
end
```
```elixir
defmodule New do
  # note that we assume the mask integer is 32-bits and do not check
  @spec mask(binary(), integer()) :: binary()
  def mask(payload, mask_integer) do

    # 1. Allocate the binary for the mask repetitions
    payload_size = byte_size(payload)
                
    mask_repetitions = case { div(payload_size,4), rem(payload_size,4)} do
      {count_4_bytes, 0 } ->
        count_4_bytes # payload length is a multiple of 4, so the mask will be fine
      {count_4_bytes, _count_stragglers} ->
       count_4_bytes + 1 # bump up by one mask size
    end
        
    mask_binary = :binary.copy( <<mask_integer::32>>, mask_repetitions)
              
    # 2. Trim the binary if needed
    fit_mask = :binary.part(mask_binary, 0, payload_size)
        
    # 3. XOR (in a nif) the payload and mask binary
    masked_payload = :crypto.exor(payload, fit_mask)
  end
end
```
```elixir
input_sm = String.duplicate("a", 10_000)
input_md = String.duplicate("a", 200_000)
input_lg = String.duplicate("a", 10_000_000)

Benchee.run(
  %{
    "original" => fn input -> Original.mask(input, 1234) end,
    "proposed" => fn input -> New.mask(input, 1234) end
  },
  inputs: %{
    "sm" => input_sm,
    "md" => input_md,
    "lg" => input_lg,
  },
  time: 5,
  memory_time: 2
)
```
```
$ mix run bench/mask.exs

Operating System: macOS
CPU Information: Apple M2 Ultra
Number of Available Cores: 24
Available memory: 128 GB
Elixir 1.15.7
Erlang 26.1.2

warmup: 2 s
time: 5 s
memory time: 2 s
reduction time: 0 ns
parallel: 1
inputs: lg, md, sm
Estimated total run time: 54 s

Benchmarking original with input lg ...
Benchmarking original with input md ...
Benchmarking original with input sm ...
Benchmarking proposed with input lg ...
Benchmarking proposed with input md ...
Benchmarking proposed with input sm ...

##### With input lg #####
Name               ips        average  deviation         median         99th %
proposed         62.47       16.01 ms     ±2.43%       15.98 ms       16.92 ms
original         19.96       50.10 ms     ±8.09%       49.28 ms       58.74 ms

Comparison:
proposed         62.47
original         19.96 - 3.13x slower +34.09 ms

Memory usage statistics:

Name        Memory usage
proposed        0.152 MB
original        43.89 MB - 287.84x memory usage +43.74 MB

##### With input md #####
Name               ips        average  deviation         median         99th %
proposed        3.20 K      312.60 μs     ±5.73%      307.96 μs      373.50 μs
original        1.73 K      576.40 μs    ±11.81%      574.40 μs      767.73 μs

Comparison:
proposed        3.20 K
original        1.73 K - 1.84x slower +263.80 μs

Memory usage statistics:

Name        Memory usage
proposed         2.88 KB
original       857.89 KB - 297.59x memory usage +855.01 KB

##### With input sm #####
Name               ips        average  deviation         median         99th %
proposed       63.83 K       15.67 μs    ±26.38%          15 μs       30.88 μs
original       29.06 K       34.42 μs   ±193.00%       20.38 μs      395.46 μs

Comparison:
proposed       63.83 K
original       29.06 K - 2.20x slower +18.75 μs

Memory usage statistics:

Name        Memory usage
proposed        0.117 KB
original        45.66 KB - 389.60x memory usage +45.54 KB

```

ryanwinchester:
Attempt No. 2

```elixir
input = String.duplicate("a", 1000)

Benchee.run(
  %{
    "original" => fn ->
      for _ <- 1..100_000, do: Original.mask(input, 1234)
    end,
    "proposed" => fn ->
      for _ <- 1..100_000, do: New.mask(input, 1234)
    end
  },
  time: 5,
  memory_time: 2
)
```

```
$ mix run bench/mask.exs
warning: variable "masked_payload" is unused (if the variable is not meant to be used, prefix it with an underscore)
  bench/mask.exs:62: New.mask/2

Operating System: macOS
CPU Information: Apple M2 Ultra
Number of Available Cores: 24
Available memory: 128 GB
Elixir 1.15.7
Erlang 26.1.2

warmup: 2 s
time: 5 s
memory time: 2 s
reduction time: 0 ns
parallel: 1
inputs: none specified
Estimated total run time: 18 s

Benchmarking original ...
Benchmarking proposed ...

Name               ips        average  deviation         median         99th %
proposed          4.80      208.30 ms     ±3.07%      207.93 ms      221.18 ms
original          2.10      475.64 ms     ±1.37%      477.59 ms      483.00 ms

Comparison:
proposed          4.80
original          2.10 - 2.28x slower +267.34 ms

Memory usage statistics:

Name        Memory usage
proposed        12.97 MB
original       534.82 MB - 41.24x memory usage +521.85 MB

```

crertel:
So, uh, that's worth considering then? :)

lib/bandit/websocket/frame.ex
```diff
@@ -184,28 +184,26 @@ defmodule Bandit.WebSocket.Frame do
...
+        
+    mask_binary = :binary.copy( <<mask_integer::32>>, mask_repetitions)
+              
+    # 2. Trim the binary if needed
+    fit_mask = :binary.part(mask_binary, 0, payload_size)
+        
+    # 3. XOR (in a nif) the payload and mask binary
+    masked_payload = :crypto.exor(payload, fit_mask)
   end
```
ryanwinchester, line 208: There are a few ways we could format this a bit more conventionally.

Here's one

```suggestion
  # Note that we assume the mask integer is 32-bits and do not check.
  @spec mask(binary(), integer()) :: binary()
  def mask(payload, mask_integer) do
    payload_size = byte_size(payload)
    count_4_bytes = div(payload_size, 4)

    mask_repetitions = if rem(payload_size, 4) == 0, do: count_4_bytes, else: count_4_bytes + 1

    fit_mask =
      <<mask_integer::32>>
      |> :binary.copy(mask_repetitions)
      |> :binary.part(0, payload_size)

    :crypto.exor(payload, fit_mask)
  end
```

mtrudel:
CI failures on missing function

mtrudel:
I find the results hard to believe (less memory usage even though we're doing one giant allocation?), but IMO it's a huge win! I wonder where else we'd benefit from this approach.

If we can get this cleaned up & it bears fruit in benchmark CI I'd be happy to merge it!

ryanwinchester:
> I find the results hard to believe (less memory usage even though we're doing one giant allocation?)

Yes, investigating the error, function name changes messed me up and my benchmarks are a giant flub. Fixing and redoing them.

lib/bandit/websocket/frame.ex
crertel: That's pretty much exactly the next iteration I think makes sense!

ryanwinchester: I think you need to also remove lines 174-185.

lib/bandit/websocket/frame.ex
```diff
@@ -184,28 +184,19 @@ defmodule Bandit.WebSocket.Frame do
     |> IO.iodata_to_binary()
   end
```
ryanwinchester, line 185: I think you need to also remove lines 174-185.

crertel:
@mtrudel :

> I find the results hard to believe (less memory usage even though we're doing one giant allocation?), but IMO it's a huge win! I wonder where else we'd benefit from this approach.

Something to keep in mind is that due to a lot of the way we use recursion in BEAM applications, and the way that immutability is enforced, it is _extremely easy_ to generate a lot of trash that the GC will have to take care of. Additionally, for certain operations--like bulk XORing here--the amount of extra code that has to get run in order to get down to effectively a single byte is nontrivial, and frequently causes allocations all along the way.

In this particular case, the C implementation in the ERTS for `exor` [here](https://github.com/erlang/otp/blob/maint/lib/crypto/c_src/math.c) is about as fast as we can hope for, without dropping into usage of SIMD--I think they're just kinda trusting the compiler vectorize that intelligently, which is a crapshoot.

There are probably other places we could benefit from this, but I'm not sure if as many of them are going to be as easy to get a big win--this is a case where the work boils down to allocating a buffer, splatting a value across it, and then doing a bulk XOR, which is about as good an argument for native acceleration as I could ask for.

Places to probably look:

* Better use of binary operations that are NIFs where it makes sense (if not already handled)
* Any place where we copy parts of binaries out that could be better handled as subbinaries (see the docs for `:binary.part`)

I'll warn though that this kind of stuff can make the code harder to read and follow, and also more brittle to change. This case was a tightly-scoped routine, but that's not going to be all cases. :)

crertel:
@ryanwinchester should be good now, removed the last of the old mask code (but kept the helpful note about involution).

mtrudel:
Thanks for the PR @crertel ! Great work!

PR #278, opened 2023-12-22
Further optimize websocket frame mask

dmorneau:
Late to the party, but suggesting a further optimization of #272 that avoids constructing the extra mask binary. With JIT improvements like https://github.com/erlang/otp/pull/5849 and https://github.com/erlang/otp/pull/7828 this could get better in future releases too.

On my system the benchmark from https://github.com/mtrudel/bandit/pull/272#issuecomment-1847571663 gives this ("proposed" being this PR):

```
Elixir 1.15.7
Erlang 26.2

warmup: 2 s
time: 5 s
memory time: 2 s
reduction time: 0 ns
parallel: 1
inputs: lg, md, sm
Estimated total run time: 54 s

Benchmarking PR272 with input lg ...
Benchmarking PR272 with input md ...
Benchmarking PR272 with input sm ...
Benchmarking proposed with input lg ...
Benchmarking proposed with input md ...
Benchmarking proposed with input sm ...

##### With input lg #####
Name               ips        average  deviation         median         99th %
proposed         41.59       24.04 ms     ±0.82%       24.02 ms       24.70 ms
PR272            38.93       25.69 ms     ±0.81%       25.64 ms       26.52 ms

Comparison:
proposed         41.59
PR272            38.93 - 1.07x slower +1.65 ms

Memory usage statistics:

Name        Memory usage
proposed        0.125 KB
PR272          249.88 KB - 1999.06x memory usage +249.76 KB

##### With input md #####
Name               ips        average  deviation         median         99th %
proposed        1.53 K      652.61 μs     ±1.86%      649.13 μs      696.49 μs
PR272           1.28 K      779.08 μs     ±2.03%      778.58 μs      824.66 μs

Comparison:
proposed        1.53 K
PR272           1.28 K - 1.19x slower +126.47 μs

Memory usage statistics:

Name        Memory usage
proposed        0.125 KB
PR272            7.72 KB - 61.75x memory usage +7.59 KB

##### With input sm #####
Name               ips        average  deviation         median         99th %
proposed        1.24 M        0.81 μs  ±1723.26%        0.75 μs        0.92 μs
PR272           0.90 M        1.11 μs  ±2424.93%        0.92 μs        2.04 μs

Comparison:
proposed        1.24 M
PR272           0.90 M - 1.38x slower +0.31 μs

Memory usage statistics:

Name        Memory usage
proposed           128 B
PR272              120 B - 0.94x memory usage -8 B

```

CC @crertel @ryanwinchester

### DIFF
--- a/lib/bandit/websocket/frame.ex
+++ b/lib/bandit/websocket/frame.ex
@@ -173,7 +173,8 @@ defmodule Bandit.WebSocket.Frame do
 
   # Note that masking is an involution, so we don't need a separate unmask function
   @spec mask(binary(), integer()) :: binary()
-  def mask(payload, mask) when is_binary(payload) and is_integer(mask) and mask >= 0x00000000 and mask <= 0xFFFFFFFF do
+  def mask(payload, mask)
+      when is_binary(payload) and is_integer(mask) and mask >= 0x00000000 and mask <= 0xFFFFFFFF do
     mask(<<>>, payload, mask)
   end
 
--- a/lib/bandit/websocket/frame.ex
+++ b/lib/bandit/websocket/frame.ex
@@ -172,19 +172,23 @@ defmodule Bandit.WebSocket.Frame do
   defp mask_and_length(length), do: <<0::1, 127::7, length::64>>
 
   # Note that masking is an involution, so we don't need a separate unmask function
-  # Note too that we assume the mask integer is 32-bits and do not check.
   @spec mask(binary(), integer()) :: binary()
-  def mask(payload, mask_integer) do
-    payload_size = byte_size(payload)
-    count_4_bytes = div(payload_size, 4)
+  def mask(payload, mask) when is_binary(payload) and is_integer(mask) and mask >= 0x00000000 and mask <= 0xFFFFFFFF do
+    mask(<<>>, payload, mask)
+  end
 
-    mask_repetitions = if rem(payload_size, 4) == 0, do: count_4_bytes, else: count_4_bytes + 1
+  defp mask(acc, <<h::32, rest::binary>>, mask) do
+    mask(<<acc::binary, (<<Bitwise.bxor(h, mask)::32>>)>>, rest, mask)
+  end
 
-    fit_mask =
-      <<mask_integer::32>>
-      |> :binary.copy(mask_repetitions)
-      |> :binary.part(0, payload_size)
+  for size <- [24, 16, 8] do
+    defp mask(acc, <<h::unquote(size)>>, mask) do
+      <<mask::unquote(size), _::binary>> = <<mask::32>>
+      <<acc::binary, (<<Bitwise.bxor(h, mask)::unquote(size)>>)>>
+    end
+  end
 
-    :crypto.exor(payload, fit_mask)
+  defp mask(acc, <<>>, _mask) do
+    acc
   end
 end


lib/bandit/websocket/frame.ex
```diff
@@ -172,19 +172,23 @@ defmodule Bandit.WebSocket.Frame do
   defp mask_and_length(length), do: <<0::1, 127::7, length::64>>
 
   # Note that masking is an involution, so we don't need a separate unmask function
-  # Note too that we assume the mask integer is 32-bits and do not check.
   @spec mask(binary(), integer()) :: binary()
-  def mask(payload, mask_integer) do
-    payload_size = byte_size(payload)
-    count_4_bytes = div(payload_size, 4)
+  def mask(payload, mask) when is_binary(payload) and is_integer(mask) and mask >= 0x00000000 and mask <= 0xFFFFFFFF do
```
ryanwinchester, line 176: do `is_binary(payload) and is_integer(mask)` add any performance benefits?
at least the integer comparisons on `mask` will fail if it is not an integer, anyway

ryanwinchester: okay the microbenchmarking i'm doing the extra guards range from not making a difference to slightly helping, so, nice.

dmorneau: It saves some work by doing the checks once in mask/2, rather than on every iteration of mask/3, similar to the examples in https://www.erlang.org/blog/more-optimizations/.

For example without checks the main `bxor` generates this bytecode:

```erlang
    {line,[{location,"lib/bandit/websocket/frame.ex",181}]}.
    {gc_bif,'bxor',
            {f,0},
            4,
            [{tr,{x,1},{t_integer,{0,4294967295}}},{x,2}],
            {x,1}}.
```

which yields arm64 with type checks and branches:

```asm
# i_bxor_jIssd
    eor x0, x26, x27
    orr x0, x0, 15
    and x8, x26, x27
    and x8, x8, 15
    cmp x8, 15
    b.eq L307
    mov x1, x26
    mov x2, x27
    bl L309
L307:
    mov x26, x0
```

Now with the checks, the compiler knows that register `{x,2}` is always a small integer:

```erlang
    {line,[{location,"lib/bandit/websocket/frame.ex",181}]}.
    {gc_bif,'bxor',
            {f,0},
            4,
            [{tr,{x,3},{t_integer,{0,4294967295}}},
             {tr,{x,2},{t_integer,{0,4294967295}}}],
            {x,3}}.
```

so it can directly xor it:

```asm
# i_bxor_jIssd
# skipped test for small operands because they are always small
    eor x28, x28, x27
    orr x28, x28, 15
```

I think the difference is not huge, maybe because the bytestring append dominates, but I hope that only gets better :)

(I used https://github.com/michalmuskala/decompile to dump bytecode, and `elixir --erl "+JDdump true" lib/bandit/websocket/frame.ex` to get JIT assembly)

crertel:
Oh, this is super neat!

I'm kinda surprised this works, given that we're pulling out the bulk binary xoring operations and are doing everything with one-off XORs.

dmorneau:
The crypto NIF is called in batches of 20kB to allow yielding ([do_exor](https://github.com/erlang/otp/blob/2bcabe9e94d6328767729d8ff4def9bc229955ab/lib/crypto/src/crypto.erl#L2380)), so on larger inputs we might actually beat it in allocations, because the append doubles the binary size every time it's full (up to 16MB, after which it grows in 20% increments).

crertel:
Ah, that would explain it. Iiiiinteresting.

mtrudel:
Missing `mix format` and this is good to go!

mtrudel:
Thanks for the PR and the patience!

PR #282, opened 2023-12-29
Support `plug: &fun/1`

wojtekmach:
There was a clause:

```elixir
plug_fn when is_function(plug_fn) -> {plug_fn, []}
```

which supported any function but in reality it only allowed 2-arity functions.

This patch adds support for 1-arity functions too. This matches `Plug.run/3`:

```elixir
iex> conn = Plug.Test.conn(:get, "/")
iex> plug = fn conn -> Plug.Conn.send_resp(conn, 200, "hi") end
iex> Plug.run(conn, [plug]).resp_body
"hi"
```

as well as projects along the lines of:

```elixir
Bypass.stub(bypass, "GET", "/", fn conn -> ... end)
```

(wink wink)

### DIFF
--- a/lib/bandit.ex
+++ b/lib/bandit.ex
@@ -20,7 +20,12 @@ defmodule Bandit do
   @typedoc """
   Possible top-level options to configure a Bandit server
 
-  * `plug`: The Plug to use to handle connections. Can be specified as `MyPlug` or `{MyPlug, plug_opts}`
+  * `plug`: The Plug to use to handle connections. Can be specified as:
+      *  `MyPlug`
+      * `{MyPlug, plug_opts}`
+      * `{&fun/2, plug_opts}`
+      * `&fun/2`
+      * `&fun/1`
   * `scheme`: One of `:http` or `:https`. If `:https` is specified, you will also need to specify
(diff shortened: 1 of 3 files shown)


mtrudel:
The Plug spec [specifically says arity 2](https://hexdocs.pm/plug/Plug.html) only; is there also a gap in Plug's docs in this respect?

wojtekmach:
I think that’s a gap yeah, I’ll send docs update and check back here.

wojtekmach:
I don't think the docs update in that spot would be useful because I believe the vast majority of Plug usage is in the form of:

```
plug Mod
plug :fun
plug :fun, opts
```

and these need to be 2-arity.

As mentioned above, &fun/1 is supported by Plug.run so I think this is legit, however mentioning it in that spot would just confuse things IMO.


I understand if you'd rather not pursue this, I'm fine doing `fn conn, _ ->` or the `fn conn, _ -> f.(conn) end` trick in my own code.

wojtekmach:
FWIW I opened https://github.com/elixir-plug/plug/issues/1201.

mtrudel:
Yep, looks like this all got squared away correctly. Thanks for the effort!

PR #289, opened 2024-01-11
Do not send a fallback if the plug has already sent a response

### LINKED ISSUE #288: Do not write 500 if an exception is already handled
Today, Bandit sends a [500 response](https://github.com/mtrudel/bandit/blob/feb275ef7422202a6ff188de5e987e0c7cbc21bc/lib/bandit/http1/handler.ex#L50) when an exception is rescued. However, when using a plug such as `Plug.ErrorHandler`, where an exception is handled and then re-raised, Bandit ends up sending a response after a response has already been written by the error handling function.

Perhaps this isn't too problematic, but I have found that some HTTP clients, such as Go's, will complain about an unexpected response being written to the open socket, with an error such as:

```text
2024/01/11 11:59:51 Unsolicited response received on idle HTTP channel starting with HTTP/1.1 500 Internal Server Error
date: Thu, 11 Jan 2024 16:59:51 GMT
content-length: 0
vary: accept-encoding
```

I couldn't track down where the exception handling happens in Plug.Cowboy, but note that when using it, this issue doesn't appear to occur.

mtrudel:
Fixes #288

### DIFF
--- a/lib/bandit/http1/handler.ex
+++ b/lib/bandit/http1/handler.ex
@@ -4,6 +4,8 @@ defmodule Bandit.HTTP1.Handler do
 
   use ThousandIsland.Handler
 
+  @already_sent {:plug_conn, :sent}
+
   @impl ThousandIsland.Handler
   def handle_data(data, socket, state) do
     connection_span = ThousandIsland.Socket.telemetry_span(socket)
@@ -109,9 +111,17 @@ defmodule Bandit.HTTP1.Handler do
   defp code_for_reason(_), do: 400
 
   defp attempt_to_send_fallback(req, code) do
-    Bandit.HTTP1.Adapter.send_resp(req, code, [], <<>>)
-  rescue
-    _ -> :ok
+    receive do
+      @already_sent ->
+        send(self(), @already_sent)
+    after
+      0 ->
+        try do
+          Bandit.HTTP1.Adapter.send_resp(req, code, [], <<>>)
+        rescue
+          _ -> :ok
+        end
+    end
   end
 
   defp maybe_keepalive(req, state) do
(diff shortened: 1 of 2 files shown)


test/bandit/http1/request_test.exs
```diff
@@ -1393,6 +1393,20 @@ defmodule HTTP1RequestTest do
       raise "boom"
     end
 
+    @tag capture_log: true
+    test "does not send an error resposne if the plug has already sent one before raising",
```
ryanwinchester, line 1397: typo in resposne

lib/bandit/http1/handler.ex
```diff
@@ -109,9 +111,17 @@ defmodule Bandit.HTTP1.Handler do
...
+      @already_sent ->
+        send(self(), @already_sent)
+    after
+      0 ->
+        try do
+          Bandit.HTTP1.Adapter.send_resp(req, code, [], <<>>)
+        rescue
+          _ -> :ok
+        end
+    end
```
ryanwinchester, line 124: it's such an obvious solution but I never would have though of this 🤔 💡 👍

test/bandit/http1/request_test.exs
mtrudel:
```suggestion
    test "does not send an error response if the plug has already sent one before raising",
```

lib/bandit/http1/handler.ex
mtrudel: Cribbed directly from plug/error_handler.ex source: https://github.com/elixir-plug/plug/blob/v1.15.2/lib/plug/error_handler.ex#L103

Issue #294, opened 2024-02-01
Request Line Read Errors


andyleclair:
Hi!
First off, thank you for Bandit, I really like using it for my personal projects!

However, we at Appcues tried to drop Bandit in as a replacement for Cowboy in our main (Phoenix 1.7) production app, but we immediately noticed a flood of crashes that we never saw under Cowboy.

All of the failures look like this:

```
GenServer #PID<0.25584.4> terminating
** (stop) "request line read error: \"04.197.85.152\\r\\n\""
Last message: {:tcp, #Port<0.4180>, "04.197.85.152\r\nX-Forwarded-Proto: https\r\nX-Forwarded-Port: 443\r\nHost: api.appcues.com\r\nX-Amzn-Trace-Id: Root=1-65bbbdc5-59ea15af3cb64254080c8a58\r\nContent-Length: 0\r\nAuthorization: Basic REDACTED\r\nAccept: application/json, text/json, text/x-json, text/javascript, application/xml, text/xml\r\nUser-Agent: RestSharp/108.0.1.0\r\ntraceparent: 00-29a1491680b1d6a4e28bea9336cfaf05-d177b554a1ca7845-00\r\nAccept-Encoding: gzip\r\nContent-Type: application/ndjson\r\n\r\n"}
```

Happy to help troubleshoot and investigate! I've got hundreds of Sentries with errors like these if more context would be helpful. It's also entirely possible that these requests are well and truly bogus and cowboy was just ignoring them, but I'm not sure. It appears to be some sort of encoding issue?

FWIW We're using Elixir 1.14.1 and OTP-25

mtrudel:
It's *possible* that you were seeing these all the time with Cowboy, since it generally does everything silently (this has been the source of innumerable issues here, with people now seeing errors things that were previously silently ignored). 

But that doesn't look like what's happening here. What you're showing looks an awful lot like Bandit attempting to start parsing an otherwise valid HTTP/1 request halfway through (specifically, the text there is obviously headers, but Bandit is expecting to be seeing the `GET /path HTTP/1.1` line that starts off all HTTP requests). There could be a number of reasons for this. Some questions:

1. What version of Bandit & Thousand Island is this (from you mix.lock file)
2. Can you provide the context of the logs for a few seconds before this?
3. If you merge the following bit of config into your phoenix config:
```
http(s): [
  http1_options: [max_requests: 1],
]
```
does the problem go away?

mtrudel:
[...]

Issue #260, opened 2023-11-05
Request Line Read Error


beckler:
I recently upgraded a code base to use Bandit, but I immediately ran into issues handling webhooks from Stripe.
While I know this isn't product interaction could be considered irrelevant, I'm unable to reproduce this same behavior with Cowboy, and I'm unable to introspect as to what exactly is going on.

I believe this is ultimately a similar issue to #252, but I'm not sure how to handle this.

Here's the results I get using Bandit mocking a `payment_intent.succeeded` event:
<img width="747" alt="Screenshot 2023-11-04 at 8 00 27 PM" src="https://github.com/mtrudel/bandit/assets/3254069/ce1d0d20-a3a1-4eae-8191-1eda41cf0349">
<img width="781" alt="Screenshot 2023-11-04 at 8 00 13 PM" src="https://github.com/mtrudel/bandit/assets/3254069/f1ed32b0-2548-4860-9a84-7f91906b6c80">

Same results using Cowboy:
<img width="782" alt="Screenshot 2023-11-04 at 7 59 26 PM" src="https://github.com/mtrudel/bandit/assets/3254069/e079a5d5-017e-4310-adc2-e10ca8ddf9cb">
<img width="781" alt="Screenshot 2023-11-04 at 7 59 28 PM" src="https://github.com/mtrudel/bandit/assets/3254069/7542c98c-dcbd-44da-ae47-f0cf0f710966">

I created a [minimal example project that demonstrates this issue](https://github.com/beckler/bandit_stripe_issue/). Granted, you'll need a Stripe account and the Stripe CLI tool to setup this test to reproduce it.

If you have any guidance here, let me know! I would love to keep using Bandit, but I don't know if I can move forward with it at the moment.

mtrudel:
Thanks for the report! 

This does look super odd; at first glance it doesn't look *too* much like #252 to my eye; that issue is about how Bandit handles requests which are obviously not HTTP (ie: intentionally very noisily), whereas this looks like we're just not parsing a (presumably otherwise valid) response properly. 

Providing a minimal repro case gets us 95% of the way there; thanks so much for that! I'll look at this first thing in the morning!

mtrudel:
I've got a repro running locally - I have a good idea what it is but it'll have to wait until morning (UTC-4 (or UTC-5 tomorrow morning!))

alisinabh:
Sorry to interrupt @mtrudel. I was looking into this and saw your comment after I found the issue!

[...]

PR #295, opened 2024-02-02
POC: Store read_state in process dictionary

alisinabh:
Hi @mtrudel,

This can be a potential temporary for issues like #260 and #294. It uses process dictionary to store the `read_state` in the HTTP1 adapter. For now it handles the error and relies on the state in dictionary (which probably is not a good idea). We can also fail, close the connection and log a better message helping users with debugging their codebases.

I personally don't like using process dictionary but in this case it seemed to me like a good use-case (at least for short while to help with integration of bandit).

P.S This is just a POC and if you think it can be useful we can work on actually implementing it properly :)

### DIFF
--- a/lib/bandit/http1/adapter.ex
+++ b/lib/bandit/http1/adapter.ex
@@ -219,6 +219,8 @@ defmodule Bandit.HTTP1.Adapter do
           |> Map.update(:req_body_bytes, byte_size(to_return), &(&1 + byte_size(to_return)))
           |> Map.put(:req_body_end_time, Bandit.Telemetry.monotonic_time())
 
+        set_process_read_state(:body_read)
+
         {:ok, to_return,
          %{req | read_state: :body_read, buffer: <<>>, body_remaining: 0, metrics: metrics}}
       else
@@ -248,6 +250,8 @@ defmodule Bandit.HTTP1.Adapter do
         |> Map.put(:req_body_bytes, byte_size(body))
         |> Map.put(:req_body_end_time, Bandit.Telemetry.monotonic_time())
 
+      set_process_read_state(:body_read)
+
       {:ok, body, %{req | read_state: :body_read, buffer: buffer, metrics: metrics}}
     end
   end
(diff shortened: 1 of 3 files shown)


mtrudel:
Ah! The process dictionary is a great idea for this (maybe also `self()` messaging, as Plug does with `@plug_sent`?)

But, as it happens (and I've done a bad job of communicating this; that's on me), I'm midway through refactoring basically the entirety of both HTTP stacks to split out the transport concerns for each protocol from the shared HTTP semantic concerns (which will be shared between them once I'm done). The first step is up at #286 and describes part of the long-term plan I've got in place there. The HTTP/1 stack is up on the bench in pieces as we speak, and I hope to have it up for review in the next week or so (https://github.com/phoenixframework/phoenix/pull/5706 distracted me a bit this week otherwise it would already be up). 

Once that lands, I think this is a *great* idea. I don't know if we need to keep the whole stack in there; maybe if we track the number of mutations of conn that we expect both inside the conn itself, and also in the process dict, and then raise an error if they mismatch when you attempt to make a conn call (or when the request ends)? Like:

```
def call(conn, _opts) do
  # pdict and conn both contain `num_calls: 0`
  {:ok, body, _conn} = read_req_body(conn)
  $ pdcit contains `num_calls: 1`, conn still contains `num_calls: 0` (since we didn't rebind the newly returned one
  
  # This call will explode with a `raise "Looks like you didn't track your conns properly"` error
  put_resp(conn, 200, "OK")
end
```

We could even make this an opt-in thing via configuration to help folks track these errors down. Then our first line of defence on issues like #294 would be to enable the `track_calls` option which would enable the above.

In any case though, I think this makes the most sense to wait for until the shared semantics work lands, since we'll get it on two stacks for the price of one at that point.

WDYT?

alisinabh:
I'm pretty much aligned with all the points you raised here. And I actually love the idea to keep the number of mutations in pdict instead of a specific part of the state.

> and I've done a bad job of communicating this; that's on me
I don't believe so. It was well communicated to me at least :)

Anyways, my intention for this PR was not get get merged but just wanted to share a potential way to detect these cases as it is very confusing. Maybe you can make a new issue for this or track this in your personal todo and close this PR. (feel free to close)

Thank you for the great work. I was really happy to see bandit as THE DEFAULT choice for new phoenix projects. 🚀

ryanwinchester:
Smart 🤔

mtrudel:
Closing for hygiene; @alisinabh this is still on my radar!

PR #297, opened 2024-02-09
Refactor HTTP/1

mtrudel:
This is a long-term working branch to handle a fairly major refactor of the HTTP/1 code. Goals:

* Split HTTP/1-specific transport concerns out from higher level HTTP semantic concerns. Put the transport concerns (roughly, RFC9112 concerns) into Bandit.HTTP1.Socket, and HTTP semantic concerns (roughly, RFC9110 concerns) into Bandit.HTTP1.Adapter. Have the two talk to one another via the `Bandit.HTTPTransport` behaviour introduced in #286. 

* Get the HTTP/1 and H2 adapters looking more or less identical

A future PR will factor them together into a single implementation. This will prevent having to implement all HTTP semantic concerns separately in both stacks, as well as provide a framework for WebSocket support in HTTP/2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,10 +3,21 @@
 ### Enhancements
 
 * Complete refactor of HTTP/2. Improved process model is MUCH easier to
-  understand and yields about a 10% performance boost to HTTP/2 requests (#286)
+  understand and yields about a 10% performance boost to HTTP/2 requests (#286 /
+  #307)
 
 ### Changes
 
+* Remove `req_line_bytes`, `req_header_bytes`, `resp_line_bytes` and
+  `resp_header_bytes` from HTTP/1 request telemetry
+* Treat trailing bytes beyond the indicated content-length on HTTP/1 requests as
+  an error
+* Surface request body read timeouts on HTTP/1 requests as `{:more...}` tuples
+  and not errors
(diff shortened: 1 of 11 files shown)


mtrudel:
This stands ready for review. The goal here is solely to refactor http/1 such that the `Bandit.HTTPTransport` behaviour is in use in both stacks, so that a subsequent PR can factor up everything else. Specifically, the following are non-goals:

* Improving the internal structure of `Bandit.HTTP1.Socket`
* Improving the structure of `Bandit.HTTP1.Handler`
* Changing the protocol implementation in any non-necessary way

A provisional changelog:

* Remove `req_line_bytes`, `req_header_bytes`, `resp_line_bytes` and `rest_header_bytes` from HTTP/1 request telemetry
* Treat trailing bytes beyond the indicated content-length on HTTP/1 requests as an error
* Surface request body read timeouts on HTTP/1 requests as `{:more...}` tuples and not errors
* Re-reading a body on HTTP/1 requests that has already been read returns `{:ok, "", conn}` instead of raising a `Bandit.BodyAlreadyReadError`
* HEAD/204/304 requests via HTTP/1 to which the plug responds with chunked encoding do not have the `transfer-encoding` header set (the relevant clause in RFC9112§6.1 lists this as a MAY).
* We no longer surface socket sending errors on chunk sends in HTTP/1

PR #300, opened 2024-02-12
GitHub Pages website, with Benchmarks, and automatic deployment

sksar:
Hi, I have created and added a GitHub Pages website. 
It is a static site generator project, located in the `www` folder.

I have also added a corresponding GitHub Action called `github-pages.yml`, which benchmarks Bandit and Cowboy, and then builds the Static site, with the results. 
During the build, the readme.md and a few files from bandit repo, are copied into the static site generator too.

<br>

Before merging this pull request, GitHub pages needs to be enabled for the Bandit repo 
Note: Due to the way the site is served, it needs to be served from a subdomain, instead of a subfolder, which is the default provided by GitHub.

Here is the demo for my fork: https://bandit.sks.sh/.
[Note this URL is temporary]

<br>

The benchmark is done using the "normal" profile. [Huge profile fails because it takes longer than 6 hours]

<br>
There are a few more Ideas to add to this, which I have in mind, however those would need further discussion. Please do let me know if you have any more ideas on this.

### DIFF
--- a/.github/workflows/github-pages.yml
+++ b/.github/workflows/github-pages.yml
@@ -62,7 +62,7 @@ jobs:
         npm run sources
         npm run build
     - name: Upload Site
-      uses: actions/upload-pages-artifact@v3
+      uses: actions/upload-pages-artifact@v2
       with:
         path: "www/build"
   deploy:
--- a/.github/workflows/github-pages.yml
+++ b/.github/workflows/github-pages.yml
@@ -47,10 +47,7 @@ jobs:
     - name: Build site
       run: |
         mv http-benchmark.csv www/sources/results/data.csv
-        sed -i 's/```elixir/```code/g' README.md
-        cp README.md www/pages/index.md
-        cp CODE_OF_CONDUCT.md www/pages/code-of-conduct.md
-        cp CHANGELOG.md www/pages/changelog.md
+        ls -al
         cd www
         npm install --force
         npm run sources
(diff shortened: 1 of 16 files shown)


mtrudel:
This looks amazing! Thanks for the hard work! 

I may have a few small change requests here (just administrative things like folder names, etc), but in general this is *extremely* impressive. I've got a busy week personally; give me a few days to sort this out and I'll report back here. 

Thanks again!

sksar:
> This looks amazing! Thanks for the hard work!
> 
> I may have a few small change requests here (just administrative things like folder names, etc), but in general this is _extremely_ impressive. I've got a busy week personally; give me a few days to sort this out and I'll report back here.
> 
> Thanks again!

Okay, do let me know. I will make the necessary changes.

iamdavidjackson:
The page title for the changelog and code of conduct pages seems to be "Evidence" instead of Bandit which is on the other pages.

sksar:
> The page title for the changelog and code of conduct pages seems to be "Evidence" instead of Bandit which is on the other pages.

Yes, I have noticed that, which I would address it later, as right now further work on this is paused for the moment.

mtrudel:
I'm (reluctantly) going to close this for hygiene reasons; it seems as if there is some momentum behind benchmarking external to the Bandit project, and honestly that's a better place for this to live. I don't really have the expertise to maintain this, and it makes more sense to be arms length anyway. 

Thanks for the work on this @sksar! Hopefully it can land externally along with the work on #327

sksar:
@mtrudel It's probably better. I agree.

PR #303, opened 2024-02-13
Disable logging of unknown messages received by http/1 handler

mtrudel:
We end up unavoidably getting a lot of messages in our mailbox from phoenix long polling, and it's becoming a support burden. There's no loss of messages here, it's just a matter of not being so verbose by default.


mtrudel:
Note this is going to merge into the 1.2.x branch, since main is taken up with the changes slated to go out in 1.3.0

PR #308, opened 2024-02-18
allow http/1.0 to work with mixed-case Keep-Alive

gregors:
tldr; Is that certain clients (in this case apache bench) will send http/1.0 requests with "Keep-Alive" and not "keep-alive" and we currently don't respect that.

I noticed this issue while comparing ab to other servers which seem to handle this correctly. I originally asked about this behavior in the [elixir forums](https://elixirforum.com/t/bandit-keep-alives-not-working-in-apache-bench/61672)

After taking a look and doing some debugging I saw that we fail to lower-case the connection header prior to comparison. Instead of incurring the penalty of lower-casing the entire header I'm assuming the mixed-case pattern match is a performant solution? Strictly speaking we probably should lower-case it to be "compliant"

I did look at the `get_connection_header_keys/1` function but it didn't really look like a good fit since it returns that list. Happy to hear feedback and change anything.

### DIFF
--- a/lib/bandit/http1/adapter.ex
+++ b/lib/bandit/http1/adapter.ex
@@ -159,6 +159,7 @@ defmodule Bandit.HTTP1.Adapter do
   # `close` & `keep-alive` always means what they say, otherwise keepalive if we're on HTTP/1.1
   defp should_keepalive?(_, "close"), do: false
   defp should_keepalive?(_, "keep-alive"), do: true
+  defp should_keepalive?(_, "Keep-Alive"), do: true
   defp should_keepalive?(:"HTTP/1.1", _), do: true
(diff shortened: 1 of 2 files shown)


lib/bandit/http1/adapter.ex
```diff
@@ -159,6 +159,7 @@ defmodule Bandit.HTTP1.Adapter do
   # `close` & `keep-alive` always means what they say, otherwise keepalive if we're on HTTP/1.1
```
mtrudel, line 159:
```suggestion
  # `close` & `keep-alive` always means what they say, otherwise keepalive if we're on HTTP/1.1
  # Case insensitivity per RFC9110§7.6.1
```

PR #312, opened 2024-02-23
Show which port is in use when startup fails

jonatanklosko:
🐈


mtrudel:
Smart!

mtrudel:
Thanks for the PR Jonatan! I'll get a point release out immediately!

jonatanklosko:
Fantastic, thank you! There is definitely no pressure, since the next Livebook release is in a while anyway, whatever works for you :)

mtrudel:
1.2.3 just released!

jonatanklosko:
<3

PR #317, opened 2024-02-28
don't send the transfer-encoding: chunked header for 100..199, 204

mwhitworth:
In [RFC 7320 section 3.3.1](https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.1), it states:

> A server MUST NOT send a Transfer-Encoding header field in any response with a status code of 1xx (Informational) or 204 (No Content).

this conditional brings Bandit into line with the RFC and Cowboy


mtrudel:
Sorry for the late reply; was on holiday last week. 

Looks great! Thanks for the PR!

mtrudel:
Released as `1.3.0`

PR #319, opened 2024-03-05
Start using `:proc_lib.set_label/1` if available

nathanl:
Starting in OTP 27.0, `:proc_lib.set_label/1` allows setting labels on processes whether or not they have a registered name. These labels are shown in Observer and in the future may be shown in other tools.

Using such labels has the potential to make a process tree more comprehensible. This commit just serves as an example for getting started.

Before:
before

After (With OTP 
after
27.0-rc1):

### DIFF
--- a/lib/bandit/application.ex
+++ b/lib/bandit/application.ex
@@ -8,6 +8,10 @@ defmodule Bandit.Application do
           {:ok, pid}
           | {:error, {:already_started, pid} | {:shutdown, term} | term}
   def start(_type, _args) do
+    if function_exported?(:proc_lib, :set_label, 1) do
+      apply(:proc_lib, :set_label, ["Bandit.Application"])
+    end
+
     children = [Bandit.Clock]
     Supervisor.start_link(children, strategy: :one_for_one)
   end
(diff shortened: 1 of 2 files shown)


lib/bandit/application.ex
```diff
@@ -8,6 +8,10 @@ defmodule Bandit.Application do
           {:ok, pid}
           | {:error, {:already_started, pid} | {:shutdown, term} | term}
   def start(_type, _args) do
+    if function_exported?(:proc_lib, :set_label, 1) do
+      apply(:proc_lib, :set_label, ["Bandit.Application"])
```
nathanl, line 12: Using `apply` here to avoid the compiler warning that is issued with OTP < 27.0

mtrudel:
Oooh I like this. I've often struggled with identifying the various parts of a Thousand Island process tree and this would be handy there as well!

nathanl:
Yeah I was thinking about your talk where you described using (IIRC) acceptor and handler processes. Labeling those would be neat.

mtrudel:
Likely best to use this via the upstream PR you're working up?

mtrudel:
I'm going to close this for hygiene; once the upstream stuff lands we can bring this back. 

Really looking forward to this both here and in Thousand Island!

Issue #320, opened 2024-03-06
Dealing with "request line read error"


Stroemgren:
I recently switched to Bandit for my Phoenix app, which has been great for performance 🥇 

However, I get errors in Sentry that I'm not sure how to deal with:
`(Sentry.CrashError ** (exit) "request line read error: \"GET /?name=Paul Leger HTTP/1.1\\r\\n\"")`

These are coming from the HTTP1 adapter and as I understand they're from reading and parsing the incoming request which is somehow malformed?

In case I'm right, is there any way to suppress or catch these? The API is serving millions of daily requests and some are bound to be malformed.

Bandit: 1.2.3
Phoenix: 1.7.11
Elixir: 1.15.7

mtrudel:
That request *is* malformed (there's a space in the request target, see [RFC9112§3](https://datatracker.ietf.org/doc/html/rfc9112#name-request-line)).

As a general policy, Bandit is intentionally noisy about errors, including client ones (I can't begin the tell you how many issues have been raised from people switching to Bandit, only to discover that the errors they were now seeing were errors on their code that had *always* been there but silently swallowed by Cowboy's policy of silence).

If you wish to disable Bandit's noisiness, you can do so as described [here](https://github.com/mtrudel/bandit/issues/252). Note that this will swallow *all* exceptional exits, which is more or less what Cowboy does.

Hopefully this solves your problem. If I've missed something, feel free to reopen this issue or file another one. Thanks!

Stroemgren:
Thanks for the quick response @mtrudel.

I do get the point about being intentionally noisy - and following the development of Bandit I've actually come across some of these stories of people realizing bugs that have been there for a long time 😺 

I'm having a harder time understanding the philosophy in this case, where it's an obvious client error. In these cases, I'd like to just return a 400 (possibly a custom one matching the content-type) and I don't see much value in seeing it in an error log. This is a public API with high volume, so I expect that there will be malformed requests all the time.

On a more practical level - it's literally exhausting my Sentry quota in no time and I don't feel like disabling all logging.

Am I missing something? Could I maybe ask how you'd go about this situation yourself?

mtrudel:
A very fair question! The short answer is: mostly because I haven't had the time to rework this properly. I've had this on the back burner for a while, but you're correct that it IS a pain. 

I think I see a way to solve this pretty quickly. Let me reopen this issue and try and put something together this evening; are you able to deploy branches of Bandit to verify it does what you want it to?

[...]

PR #321, opened 2024-03-07
Provide config option to disable logging of protocol errors

mtrudel:
As a potential fix to #320, add an option to disable the logging of protocol errors in HTTP/1. Logging is still enabled by default (at least for now; I'd like to get some idea of how many errors this ends up hiding before enabling by default). You can disable logging by:

```
config :my_phoenix_app, MyPhoenixAppWeb.Endpoint,
  http: [
    ip: ...,
    port: ...,
    http_1_options: [log_protocol_errors: false]
  ]
  ...
```


mtrudel:
Released as `1.3.0`

Issue #313, opened 2024-02-26
Increased memory usage after switching from Cowboy to Bandit


Nilsonn:
Hey,
we recently migrated a Phoenix application from Cowboy to Bandit, following the recommendation of the Phoenix framework for future projects.
After the transition, we have observed an increase in memory consumption by our application. Previously, our application used approximately 300 MB of memory, but under Bandit we reach our peak at over 400 MB.
Correlating with the increased memory, we see an increase in processes with a peak of approximately 200 more processes.
<img width="1223" alt="Bildschirmfoto 2024-02-26 um 11 08 00" src="https://github.com/mtrudel/bandit/assets/6068179/8156f43e-631d-4146-9c12-581b5061782a">

Used Bandit Versions: 1.2.0-1.2.2 
Elixir Version: 1.16.0
Erlang: 26.2.1
OS: Ubuntu Jammy

We have not made any significant changes to our codebase other than the web server switch, and the increased memory usage is consistent across multiple deployments. Further, we did not get a lot more users.

Is the increase in memory and processes expected?

Let me know if I can provide anything else.

Best regards
Nils

mtrudel:
What version of thousand island are you using? 

What web traffic (http/weboscket) and transport (http/https) are you using?

Nilsonn:
I bumped thousand island to 1.3.5 and Bandit to 1.2.3 yesterday morning.
The memory increase still occurs and reaches the same peak values as with the previous versions.
Do you have any other ideas why this might happen with Bandit but did not with Cowboy?
<img width="1737" alt="Bildschirmfoto 2024-02-28 um 07 07 21" src="https://github.com/mtrudel/bandit/assets/6068179/74caf833-a549-496b-8563-304c341494e8">

mtrudel:
Interesting! What are the markers on those charts?

Nilsonn:
Do you mean the rocket and the dots below the charts?
Those represent deployments. The rocket deployment was the update to thousand island 1.3.5 and Bandit 1.2.3.

And the two colored lines represent two nodes that are running.

Do you need any more information to understand this better?

ianko:
@mtrudel Allow me to share the same experience as described by @Nilsonn.

We ran 3 experiments in the past couple days (30min, 30min, 6hours), where a new pod running Bandit was introduced alongside to other pods running Cowboy.

[...]

PR #322, opened 2024-03-08
Run an explicit garbage collection between every 'n' keepalive requrests in HTTP/1

mtrudel:
Solves #313 

'n' defaults to 5, but is configurable using the `gc_every_n_keepalive_requests` http_1 option.

### DIFF
--- a/lib/bandit.ex
+++ b/lib/bandit.ex
@@ -94,6 +94,9 @@ defmodule Bandit do
     Defaults to 50 headers
   * `max_requests`: The maximum number of requests to serve in a single
     HTTP/1.1 connection before closing the connection. Defaults to 0 (no limit)
+  * `gc_every_n_keepalive_requests`: How often to run a full garbage collection pass between subsequent
+    keepalive requests on the same HTTP/1.1 connection. Defaults to 5 (garbage collect between
+    every 5 requests). This option is currently experimental, and may change at any time
   * `log_unknown_messages`: Whether or not to log unknown messages sent to the handler process.
     Defaults to `false`
   * `log_protocol_errors`: Whether or not to log protocol errors such as malformed requests.
@@ -110,6 +113,7 @@ defmodule Bandit do
           max_header_length: pos_integer(),
           max_header_count: pos_integer(),
           max_requests: pos_integer(),
+          gc_every_n_keepalive_requests: pos_integer(),
           log_unknown_messages: boolean(),
           log_protocol_errors: boolean(),
           compress: boolean(),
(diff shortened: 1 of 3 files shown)


lib/bandit.ex
```diff
@@ -110,6 +113,7 @@ defmodule Bandit do
           max_header_length: pos_integer(),
           max_header_count: pos_integer(),
           max_requests: pos_integer(),
+          gc_every_n_requests: pos_integer(),
```
vanderhoop, line 116: Given this config only affects http1 keepalive requests, I would change the key name here to `gc_every_n_keepalive_requests` to avoid arresting more attention than necessary. 

As a developer, I can intuit why a gc config might be necessary for long lived keepalives, but I'd be somewhat suspicious of a config for any and all http1 reqs.

mtrudel: It's already inside the http_1_options subgroup

mtrudel: Updated.

lib/bandit/http1/handler.ex
```diff
@@ -140,9 +140,16 @@ defmodule Bandit.HTTP1.Handler do
 
     if under_limit && req.keepalive do
       case ensure_body_read(req) do
-        :ok -> {:continue, Map.put(state, :requests_processed, requests_processed)}
-        {:error, :closed} -> {:close, state}
-        {:error, reason} -> {:error, reason, state}
+        :ok ->
+          gc_every_n_requests = Keyword.get(state.opts.http_1, :gc_every_n_keepalive_requests, 5)
+          if rem(requests_processed, gc_every_n_requests) == 0, do: :erlang.garbage_collect()
+          {:continue, Map.put(state, :requests_processed, requests_processed)}
```
vanderhoop, line 146: __Inner monologue:__ "Could we reproduce, via test, the memory bloat described in #313 with a sequence of keepalive requests over a single connection? And then assert that the memory is reduced about N of those requests?"

vanderhoop: ☝️ To be clear, not implying you need a test before merge. Just wondered if the above would be possible.

vanderhoop: I think it'd have to be in an `async: false` test to isolate things, but maybe even then it wouldn't be possible?

mtrudel: Yeah...... I was thinking the same thing. I can likely dig into the Erlang statistics call and validate. lemme look into it

mtrudel:
Just waiting to hear back from @ianko about the validity of the default value of `5` for `gc_every_n_keepalive_requests` and then this is good merge & release.

ianko:
@mtrudel sorry about the wait. I will start a new experiment in about 30min. Will share as soon I get the results.

mtrudel:
No worries no hurries mate! Take your time.

lib/bandit/http1/handler.ex
vanderhoop: > there's no explicit way to get a count / notice of full sweep GCs that I can find, so I've gotta resort to side effects

I think that's better, as the side effect is the important thing, right?

And if you can reliably red/green it by changing the config, 🤌. Great work!

ianko:
Shared metrics on https://github.com/mtrudel/bandit/issues/313#issuecomment-1986510150 for better comparison in a single place.

This experiment run had some abnormal requests, which made the latency `max` metrics to jump. But it also happened to other pods running Cowboy.

I can run another longer experiment on Monday, if needed.

Issue #345, opened 2024-04-24
High CPU usage


MrYawe:
First, thanks for all the amazing work on Bandit :)

In our production we had CPU load increase over time and we fixed it the 22/04 by switching back to Cowboy.
We were using bandit 1.4.2 and thousand_island 1.3.5.

<img width="1168" alt="image" src="https://github.com/mtrudel/bandit/assets/8686039/92e31546-6f48-4128-96d3-7e220a73e578">
<img width="1168" alt="image" src="https://github.com/mtrudel/bandit/assets/8686039/afe7be1e-4f0e-4167-825d-dad227b5976c">

Last weekend (20/04-21/04), I took some screenshots about our live running processes.

We can see that processes with highest reductions count (CPU usage) are all LiveView processes:
<img width="1197" alt="image" src="https://github.com/mtrudel/bandit/assets/8686039/55b9a3e8-4e6b-418d-b238-6c2a247d4d95">

`Bandit.DelegatingHandler.init/1` was often in the top most memory-hungry processes. Maybe there is a link with https://github.com/mtrudel/bandit/issues/339:

<img width="1197" alt="image" src="https://github.com/mtrudel/bandit/assets/8686039/b21db168-98f3-4635-9852-c0b593912aeb">

PS: these two LiveDashboard screenshots were not taken at the same time

mtrudel:
I have a hunch that this may be due to older connections not shutting down when the client disconnects (which traces back to the same issues as #344). It's actually next up on my list of things to look at!

mtrudel:
Update: It's not the same issue as #344, but I've got a repro case for this locally (it looks like it may be related to what's left of #339). Not entirely sure what's causing it yet, but I can reproduce a significant reduction / memory difference between Cowboy and Bandit on the exact same workload, so it's just a matter of digging until I can explain / mitigate it.

mtrudel:
Closing due to inactivity

Issue #202, opened 2023-07-27
Stopping the request handler


hubertlepicki:
I have two issues, that are slightly related.

I have noticed that if I leave this line in my Phoenix controllers, the process that handles request never times out, never crashes, it just continues to exit seemingly forever:

```
def index(conn, params) do
  :timer.sleep(:infinity)
  ...
end
```

The process continues to remain alive even if the client browser closes.

The two problems I have with the above behavior are:

1. The request handler does not have a timeout set, and I can't seem to find in Bandit's documentation how to set such timeout. Ideally, I'd like the request handler to be killed after X seconds of not completing it's job - is there a way for me to do it? I only found `:read_timeout` option which is useful but not in this particular case, as I can see from the source of this project that it only gets applied to the time the request body is being read, not to the request handler itself.

2. Is there a way to immediately kill the request handler when the client disconnects (i.e. browser tab closes)?

I have tried adding the following config, but it seems to have no effect either:

```
 config :ui, UI.Endpoint,
   http: [
-    http_1_options: [max_request_line_length: 12288, max_header_length: 12288, compress: true]
+    http_1_options: [max_request_line_length: 12288, max_header_length: 12288, compress: true],
+    thousand_island_options: [read_timeout: 10_000]
   ],
   adapter: Bandit.PhoenixAdapter,
  ...
```

hubertlepicki:
I did some digging and it appears that this is simply not implemented, and I am not sure if that's intentional or not.

@mtrudel I could possibly add the missing functionality if you agree and help me out a bit with directions.

As of 1) the info I would love to hear from you is if it should be handled within Bandit at all, because this can be done on a Plug level. I looked at Cowboy ant it doesn't appear to have such setting either, so my guess is that we don't want it here either.

When it comes to 2) - the ThousandIsland.Handler behavior specifies `handle_close`, which is used in the `Bandit.Websocket.Handler` here https://github.com/mtrudel/bandit/blob/31dbb2f6caed7d0bedfdb566888a30c8b0d125b1/lib/bandit/websocket/handler.ex#L72 but not in (as far as I can tell) Http1 or Http2 handlers.

And even if added, I think the callback would *not* interrupt the above example, with Phoenix controller (or any Plug, really) sleeping to infinity. What I suspect would happen, instead, becaues Handler is a GenServer, is that the messge will be delivered only after the Plug stops working, and it will never interrupt the process that sleeps forever.

I suspect this may be the same issue with Websocket handler.

[...]

PR #346, opened 2024-05-03
Disable exit trapping for the duration of HTTP/1 requests 

### LINKED ISSUE #344: Plug.Conn is not closed when user closes the browser's tab (SSE)
Hi,
I have implemented SSE (Server-Sent-Events) using Plug and Bandit, and they work really cool until a user closes browser's tab. For some reason, my Plug.Conn isn't notified/invalidated and I can still send data (although it doesn't appear in Wireshark).  The code looks more or less like this

```elixir
 get "/api/resource/:resource_id/sse/event-stream" do
    viewers = PeerConnection.get_all_running() |> length()

    {:ok, conn} =
      conn
      |> put_resp_header("content-type", "text/event-stream")
      |> put_resp_header("cache-control", "no-cache")
      |> send_chunked(200)
      |> chunk("data: {\"viewerscount\": #{viewers}}\n\n")

    update_viewers_count(conn, self())
  end


  defp update_viewers_count(conn, pid) do
    viewers = PeerConnection.get_all_running() |> length()
    Process.send_after(self(), :update_viewerscount, @viewers_count_update_interval)

    receive do
      :update_viewerscount ->
        {:ok, conn} = chunk(conn, "data: {\"viewerscount\": #{viewers}}\n\n")
        update_viewers_count(conn, pid)
    end
  end
```

When I switched to the cowboy, everything works as expected, i.e. once the tab is closed, a process handling a connection on the server side is also closed.

mtrudel:
Fixes #344 #202, maybe #345

### DIFF
--- a/lib/bandit/http1/handler.ex
+++ b/lib/bandit/http1/handler.ex
@@ -6,6 +6,10 @@ defmodule Bandit.HTTP1.Handler do
 
   @impl ThousandIsland.Handler
   def handle_data(data, socket, state) do
+    # Plug calls may take arbitrarily long, and since they all happen inside the current stack we
+    # have no way to be interrupted if the client closes the request other than to stop trapping
+    # exits. Requests that are closed mid-request will thus eagerly terminate the server process
+    Process.flag(:trap_exit, false)
     transport = %Bandit.HTTP1.Socket{socket: socket, buffer: data, opts: state.opts}
     connection_span = ThousandIsland.Socket.telemetry_span(socket)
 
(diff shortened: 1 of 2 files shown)


mtrudel:
In principle this seems pretty uncontroversial; need to add a few tests here to make sure this is doing what I think it is...

mtrudel:
This doesn't actually work the way I thought it did - exit trapping doesn't have any effect on whether the socket closes the handler process.

PR #348, opened 2024-05-10
Run GC when switching to websocket

mtrudel:


mtrudel:
I think this is pretty uncontroversial; I'm merging as soon as CI comes back green

PR #351, opened 2024-05-10
Only save Process keys that are atoms starting with $

### LINKED ISSUE #350: Clearing process dictionary breaks on some keys
ref: https://github.com/phoenixframework/phoenix/issues/5807

ryanwinchester:
Fixes #350

### DIFF
--- a/lib/bandit/http1/handler.ex
+++ b/lib/bandit/http1/handler.ex
@@ -33,8 +33,12 @@ defmodule Bandit.HTTP1.Handler do
   end
 
   defp clear_process_dict do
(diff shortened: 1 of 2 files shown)


mtrudel:
Looks great! Can we add a line like https://github.com/mtrudel/bandit/blob/main/test/bandit/http1/request_test.exs#L252 that adds a tuple term to the dictionary

ryanwinchester:
Updated, but I have to head out for a bit. If you need to take over or redo, please feel free

mtrudel:
1.5.2 coming momentarily

PR #353, opened 2024-05-13
Respect pre-computed content-length for HEAD responses

meeq:
I have an application that streams extremely large payloads. Due to reasons beyond my control, HEAD requests must respond with HTTP status 200, so my Plug handler sends with:

```elixir
conn
|> Plug.Conn.put_resp_header("content-length", payload_size)
|> Plug.Conn.send_resp(200, "")
```

However, `Bandit.Headers.add_content_length/3` replaces my application's "content-length" header with `content-length: 0`. 

As it stands the only way for me to set "content-length" on a HEAD 200 response seems to be to either wastefully render the full response or allocate a body of "content-length"-size and pass it into `send_resp`, but this is not practical or efficient. It also seems to violate the spirit of [RFC9110 § 9.3.2](https://www.rfc-editor.org/rfc/rfc9110.html#HEAD):

>  A client SHOULD NOT generate content in a HEAD request unless it is made directly to an origin server that has previously indicated, in or out of band, that such a request has a purpose and will be adequately supported.

This patch introduces a simple, reasonably-safe heuristic to preserve the application's intent: if a HEAD response handler sets the content-length response header, treat it as the size of the body that would have been sent; [RFC9110 §8.6](https://www.rfc-editor.org/rfc/rfc9110.html#name-content-length):

>A server MAY send a Content-Length header field in a response to a HEAD request; a server MUST NOT send Content-Length in such a response unless its field value equals the decimal number of octets that would have been sent in the content of a response if the same request had used the GET method. [...] A server MUST NOT send a Content-Length header field in any response with a status code of **1xx (Informational)** or **204 (No Content)**.

### DIFF
--- a/lib/bandit/adapter.ex
+++ b/lib/bandit/adapter.ex
@@ -126,10 +126,8 @@ defmodule Bandit.Adapter do
     compress = Keyword.get(adapter.opts.http, :compress, true)
     headers = if compress, do: [{"vary", "accept-encoding"} | headers], else: headers
 
-    headers =
-      if Bandit.Headers.override_content_length?(headers, status, adapter.method),
-        do: Bandit.Headers.add_content_length(headers, IO.iodata_length(body), status),
-        else: headers
+    length = IO.iodata_length(body)
+    headers = Bandit.Headers.add_content_length(headers, length, status, adapter.method)
 
     metrics =
       adapter.metrics
@@ -159,7 +157,7 @@ defmodule Bandit.Adapter do
     length = if length == :all, do: size - offset, else: length
 
     if offset + length <= size do
-      headers = Bandit.Headers.add_content_length(headers, length, status)
+      headers = Bandit.Headers.add_content_length(headers, length, status, adapter.method)
       adapter = send_headers(adapter, status, headers, :raw)
 
       {socket, bytes_actually_written} =
--- a/lib/bandit/adapter.ex
+++ b/lib/bandit/adapter.ex
@@ -125,7 +125,11 @@ defmodule Bandit.Adapter do
 
     compress = Keyword.get(adapter.opts.http, :compress, true)
     headers = if compress, do: [{"vary", "accept-encoding"} | headers], else: headers
-    headers = Bandit.Headers.add_content_length(headers, IO.iodata_length(body), status)
+
+    headers =
+      if Bandit.Headers.override_content_length?(headers, status, adapter.method),
+        do: Bandit.Headers.add_content_length(headers, IO.iodata_length(body), status),
+        else: headers
 
     metrics =
       adapter.metrics
--- a/lib/bandit/headers.ex
+++ b/lib/bandit/headers.ex
@@ -95,7 +95,7 @@ defmodule Bandit.Headers do
         ) ::
           Plug.Conn.headers()
 
-  # For HEAD responses, respect existing the content-length header (if set)
+  # For HEAD responses, respect the existing content-length header (if set)
   # EXCEPT if the status code forbids a content-length header.
   # If content-length is not set, the server will not send a content-length header.
   def add_content_length(headers, 0, status, "HEAD") do
--- a/lib/bandit/headers.ex
+++ b/lib/bandit/headers.ex
@@ -87,36 +87,41 @@ defmodule Bandit.Headers do
 
   defp parse_integer(rest, total), do: {total, rest}
 
-  @spec add_content_length(Plug.Conn.headers(), non_neg_integer(), Plug.Conn.int_status()) ::
+  @spec add_content_length(
+          headers :: Plug.Conn.headers(),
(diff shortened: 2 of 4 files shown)


lib/bandit/adapter.ex
```diff
@@ -125,7 +125,11 @@ defmodule Bandit.Adapter do
 
     compress = Keyword.get(adapter.opts.http, :compress, true)
     headers = if compress, do: [{"vary", "accept-encoding"} | headers], else: headers
-    headers = Bandit.Headers.add_content_length(headers, IO.iodata_length(body), status)
+
+    headers =
+      if Bandit.Headers.override_content_length?(headers, status, adapter.method),
+        do: Bandit.Headers.add_content_length(headers, IO.iodata_length(body), status),
+        else: headers
```
mtrudel, line 132: I'd prefer to see this inside `Bandit.Headers.add_content_length/3` as then all of the logic for this is centralized

mtrudel:
Thanks for the PR! Left you some refactoring suggestions. We'll also need a test to cover this for both H1 and H2.

lib/bandit/headers.ex
```diff
@@ -97,6 +97,22 @@ defmodule Bandit.Headers do
...
+  @spec override_content_length?(
+          headers :: Plug.Conn.headers(),
+          status :: Plug.Conn.int_status(),
+          method :: Plug.Conn.method()
+        ) ::
+          boolean()
+
+  def override_content_length?(headers, status, "HEAD") do
+    value = get_header(headers, "content-length")
+    is_nil(value) or not add_content_length?(status)
```
meeq, line 111: Upon further reflection, I actually believe that the server should only override the content-length of HEAD responses if the status code requires it to be omitted.

Specifically, the language in [RFC9110 §8.6](https://www.rfc-editor.org/rfc/rfc9110.html#name-content-length):
> A server MAY send a Content-Length header field in a response to a HEAD request; a server MUST NOT send Content-Length in such a response unless its field value equals the decimal number of octets that would have been sent in the content of a response if the same request had used the GET method.

* the server has no way of knowing what the length of the GET would have been unless it's provided as the body
* [§9.3.2](https://www.rfc-editor.org/rfc/rfc9110.html#section-9.3.2) is clear that HEAD requests should not generate and discard the content in order to populate response headers
* HEAD responses will never send the body, so it doesn't make sense to set the body to anything other than `""`
* the server is not required to set a Content-Length header for HEAD responses

When the application does not specify a Content-Length for HEAD responses, it would not be correct for the server to impose the response's Content-Length based on the size of the response body.

test/bandit/http2/plug_test.exs
```diff
@@ -214,11 +214,10 @@ defmodule HTTP2PlugTest do
     response = Req.head!(context.req, url: "/header_write_test")
 
     assert response.status == 200
-    assert map_size(response.headers) == 5
+    assert map_size(response.headers) == 4
 
     assert %{
              "date" => [date],
-             "content-length" => ["0"],
```
meeq, line 221: This evidences a "breaking change" for existing users who depend on HEAD always sending back "content-length: 0".

As mentioned in an earlier comment, the existing behavior is arguably not correct, but [it has been enshrined](https://xkcd.com/1172/).

meeq: I have [added tests](https://github.com/mtrudel/bandit/pull/353/commits/7018508b549a5ce4f3eb742e257a4577e9d844a0) to exercise how an application would explicitly set "content-length: 0" on HEAD responses:

```elixir
conn
|> put_resp_header("content-length", "0")
|> send_resp(200, "")
```

Unfortunately, I can't see a feasible way to disambiguate "incidentally zero-length" (i.e. `send_resp(200, "")`) from "intentionally zero-length" (i.e. `send_resp(200, generated_content) when byte_size(generated_content) == 0`) in the computed response body size.

mtrudel: Yep this is a fine breaking change - for one it's super subtle, and for two, it's not something anyone would get themselves except deliberately. Thanks for surfacing it though!

lib/bandit/headers.ex
```diff
@@ -87,20 +87,41 @@ defmodule Bandit.Headers do
...
+          length :: non_neg_integer(),
+          status :: Plug.Conn.int_status(),
+          method :: Plug.Conn.method()
+        ) ::
           Plug.Conn.headers()
-  def add_content_length(headers, length, status) do
-    headers = Enum.reject(headers, &(elem(&1, 0) == "content-length"))
 
-    if add_content_length?(status),
+  # For HEAD responses, respect existing the content-length header (if set)
```
mtrudel, line 98:
```suggestion
  # For HEAD responses, respect the existing content-length header (if set)
```

mtrudel (approved):
One minor typo otherwise this is good to go!

PR #359, opened 2024-05-24
Raise error on send errors for HTTP/1 responses

### LINKED ISSUE #344: Plug.Conn is not closed when user closes the browser's tab (SSE)

mtrudel:
The Plug spec doesn't really provide for the effective surfacing of errors from send calls, mostly because Cowboy will (usually) have killed off the connection handler process when the client closes the connection. In Bandit's case, we don't actively kill off connection processes in this situation, so we need to be able to surface these cases effectively. Raising a `Bandit.HTTPError` will bubble through the caller all the way up to Bandit's handler pipeline, and log the error as configured. This is as good of a pattern as we're going to have, and mirrors what we do for reads against closed clients already (tests were added for that case here as well, since they were absent before). 

Fixes #344

### DIFF
--- a/lib/bandit/adapter.ex
+++ b/lib/bandit/adapter.ex
@@ -195,7 +195,14 @@ defmodule Bandit.Adapter do
     # this entire section of the API is a bit slanty regardless.
 
     validate_calling_process!(adapter)
-    {:ok, nil, send_data(adapter, chunk, IO.iodata_length(chunk) == 0)}
+
+    # chunk/2 is unique among Plug.Conn.Adapter's sending callbacks in that it can return an error
+    # tuple instead of just raising or dying on error. Rescue here to implement this
+    try do
+      {:ok, nil, send_data(adapter, chunk, IO.iodata_length(chunk) == 0)}
+    rescue
+      error -> {:error, Exception.format_banner(:error, error)}
+    end
   end
 
   @impl Plug.Conn.Adapter
--- a/lib/bandit/adapter.ex
+++ b/lib/bandit/adapter.ex
@@ -201,7 +201,7 @@ defmodule Bandit.Adapter do
     try do
       {:ok, nil, send_data(adapter, chunk, IO.iodata_length(chunk) == 0)}
     rescue
-      error -> {:error, Exception.format_banner(:error, error)}
+      error -> {:error, error.message}
     end
   end
 
--- a/lib/bandit/http1/socket.ex
+++ b/lib/bandit/http1/socket.ex
@@ -332,15 +332,15 @@ defmodule Bandit.HTTP1.Socket do
 
         :chunk_encoded ->
           headers = [{"transfer-encoding", "chunked"} | headers]
-          _ = ThousandIsland.Socket.send(socket.socket, [resp_line | encode_headers(headers)])
+          send!(socket.socket, [resp_line | encode_headers(headers)])
           %{socket | write_state: :chunking}
 
         :no_body ->
-          _ = ThousandIsland.Socket.send(socket.socket, [resp_line | encode_headers(headers)])
+          send!(socket.socket, [resp_line | encode_headers(headers)])
           %{socket | write_state: :sent}
 
         :inform ->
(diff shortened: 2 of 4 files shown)


CharlesOkwuagwu:
>>>In Bandit's case, we don't actively kill off connection processes in this situation, so we need to be able to surface these cases

Please is there a reason why you keep the connection process alive when the client closes?

Can this behavior be a configurable option? 

For cases where you need well behaved ServerSentEvents (SSE) killing the process is the correct behavior when the client closes.

mtrudel:
> Please is there a reason why you keep the connection process alive when the client closes?

It's a side-effect of the process model we use; we don't really have a choice. The key thing is that Bandit only uses a single process per connection, and this connection uses the `active: :once` mode of the underlying socket (gen_tcp/ssl). The *only* way to be informed of the underlying connection being closed is to either 1. Receive `:tcp_closed` messages, or  2. handle `{:error, :closed}` tuples that come back from the various socket functions. This PR actively adds support for that second option on send calls.

CharlesOkwuagwu:
I've used the latest version of Bandit, I setup a test specifically for SSE. 

THE handling process stays alive, long after the client closes. 

Us this the expected behavior, for the option 2 you mentioned above?

mtrudel:
> I've used the latest version of Bandit, I setup a test specifically for SSE.

I'd love to see a repro for this! The SSE process is obviously doing *something* while it's sitting idle and it may be possible to have it look at the connection status while it's doing this

mtrudel:
(Feel free to reopen this issue, btw)

CharlesOkwuagwu:
> > I've used the latest version of Bandit, I setup a test specifically for SSE.
> 
> I'd love to see a repro for this! The SSE process is obviously doing _something_ while it's sitting idle and it may be possible to have it look at the connection status while it's doing this

Sorry for the delay

https://github.com/CharlesOkwuagwu/sse_demo

CharlesOkwuagwu:
Please open here in your web browser (MS-edge) https://localhost:9090/demo

mtrudel:
I see the same behaviour in both Chrome & Safari. Unable to test on Edge, but everything here looks like the server is doing the right thing; I wonder if it's an OS issue (I'm on macOS) or a browser issue? Clients don't always close the underlying connection right away after you close the window, instead keeping it open on a timer for possible reuse. I wonder if that could be what's happening in your case?

CharlesOkwuagwu:
> I see the same behaviour in both Chrome & Safari. Unable to test on Edge, but everything here looks like the server is doing the right thing; I wonder if it's an OS issue (I'm on macOS) or a browser issue? Clients don't always close the underlying connection right away after you close the window, instead keeping it open on a timer for possible reuse. I wonder if that could be what's happening in your case?

WOW you are right, I Was closing only the tab. When I close the browser. the process exits.

Sorry for the disturbance.

mtrudel:
No worries! Glad it all got figured out!

PR #380, opened 2024-07-11
accept hpax 1.0

rodrigues:
Hi 👋

`hpax` 1.0.0 upgrade just removes a warning when compiling in Elixir 1.17:

```elixir
warning: 62..@dynamic_table_start + length - 1 inside guards requires an explicit step, please write 62..@dynamic_table_start + length - 1//1 or 62..@dynamic_table_start + length - 1//-1 instead
  lib/hpax/table.ex:160: HPAX.Table.lookup_by_index/2
```

Thanks!

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -32,7 +32,7 @@ defmodule Bandit.MixProject do
       {:thousand_island, "~> 1.0"},
       {:plug, "~> 1.14"},
       {:websock, "~> 0.5"},
-      {:hpax, "~> 0.2.0"},
+      {:hpax, "~> 0.2.0 or ~> 1.0"},
       {:telemetry, "~> 0.4 or ~> 1.0"},
       {:req, "~> 0.3", only: [:dev, :test]},
       {:machete, ">= 0.0.0", only: [:dev, :test]},


alisinabh:
Since bandit already requires elixir 1.13+ and hpax 1.0.0 requires 1.12+ (lower) I guess we don't need to support 0.2.0 anymore?

cc @mtrudel

mtrudel:
I think you're correct @alisinabh, but the only wrinkle is that mint only added version support for hpax 1.0.0 [last week](https://github.com/elixir-mint/mint/commit/5d0c6e9aba84618a7fd374cfa49ad392d2125061), so we'd be forcing people that use mint in their apps (which is a LOT of people) to update to at least mint 1.6.2 to solve their dev tree. Not ideal.

mtrudel:
(Also ignore the busted CI; it's an issue with Req that I'm working up)

sleipnir:
> I think you're correct @alisinabh, but the only wrinkle is that mint only added version support for hpax 1.0.0 [last week](https://github.com/elixir-mint/mint/commit/5d0c6e9aba84618a7fd374cfa49ad392d2125061), so we'd be forcing people that use mint in their apps (which is a LOT of people) to update to at least mint 1.6.2 to solve their dev tree. Not ideal.

Hi @mtrudel I don't see any problem in updating because, in fact, the ideal is to stay updated, right?

PR #385, opened 2024-08-01
Expose body read timeouts as 408 Request Timeout errors

### LINKED ISSUE #384: HTTP1 socket might be overly tolerant of socket timeout error
STR:
1. Send a simple HTTP1.1 POST request to a Bandit-powered Phoenix application, but set `Content-Length` header bigger than the actual body length
2. The request returns `413 Request Entity Too Large` after waiting for `read_timeout` (default 60_000) ms.

I believe the reason for this is that Bandit just returns what [it read so far in case of socket timeout](https://github.com/mtrudel/bandit/blob/dc842949996a8069f068263abd6331834be15b3a/lib/bandit/http1/socket.ex#L316). I wonder if this is the correct behaviour here? I'm not an expert on interpreting HTTP standard, but reading through [RFC9112§6.3.6](https://www.rfc-editor.org/rfc/rfc9112.html#section-6.3-2.6) it seems like it's calling for a more brutal handling?

> If a valid [Content-Length](https://www.rfc-editor.org/rfc/rfc9112.html#body.content-length) header field is present without [Transfer-Encoding](https://www.rfc-editor.org/rfc/rfc9112.html#field.transfer-encoding), its decimal value defines the expected message body length in octets. If the sender closes the connection **or the recipient times out before the indicated number of octets are received**, the recipient MUST consider the message to be incomplete and close the connection.

`Plug.Conn.Adapter.read_req_body/2` callback support error tuple as a response, so perhaps `Bandit.Adapter` should use it instead of returning a `:more` tuple?

For context, after switching to Bandit, we began to see mysterious `Plug.Parsers.RequestTooLargeError` errors from requests that are well below default `length` limit of `8_000_000` bytes. After some investigation, I stumbled upon this behaviour. I'm not sure if this is indeed what we see and if it is, why content length would be incorrect for some requests, but I'm worried that `RequestTooLargeError` may be obfuscating real error in some cases.

mtrudel:
Fixes #384

### DIFF
--- a/lib/bandit/http1/socket.ex
+++ b/lib/bandit/http1/socket.ex
@@ -313,7 +313,7 @@ defmodule Bandit.HTTP1.Socket do
           end
 
         {:error, :timeout} ->
-          already_read
+          request_error!("Body read timeout", :request_timeout)
 
         {:error, reason} ->
           request_error!(reason)
--- a/test/bandit/http1/request_test.exs
+++ b/test/bandit/http1/request_test.exs
@@ -875,17 +875,16 @@ defmodule HTTP1RequestTest do
             "POST /short_body HTTP/1.1\r\nhost: localhost\r\ncontent-length: 5\r\n\r\nABC"
           )
 
-          assert {:ok, "200 OK", _, "OK"} = SimpleHTTP1Client.recv_reply(client)
+          assert {:ok, "408 Request Timeout", _, ""} = SimpleHTTP1Client.recv_reply(client)
           Process.sleep(1100)
         end)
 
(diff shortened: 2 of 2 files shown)


mtrudel:
This solution explicitly raises an HTTPError because returning an error tuple from `Plug.Conn.Adapter.read_req_body/2` doesn't provide us with a way to track the failed read on the connection, so there's no way to meaningfully manage the connection once this happens. This puts a lot of onus on the user to properly ensure that nothing further gets done on the connection, 

It also makes it really tricky to send a useful error to the client, and also to ensure that the connection is closed afterwards.

Moreover, it keeps the flow for this case the same as it is for all other HTTP error cases; the proper client signalling here just falls out of Bandit's design.

PR #396, opened 2024-08-30
Handle `throw`ing plug call

grzuy:
Hi,

Thanks for writing this great library!

It seems that uncaught `throw`s inside the plug call currently end up with a socket closed the following in the logs

```
[error] GenServer #PID<0.4420.0> terminating
** (stop) bad return value: "something was thrown"
```
(no stacktrace)

With the changes in this PR

```
[error] ** (throw) "something was thrown"
    ... PROPER STACKTRACE HERE ...
```

and a corresponding telemetry event, similar to exceptions.

Thoughts?

Haven't looked closely yet, but I think the same can be said of non-normal exits. Will take a look later.

### DIFF
--- a/lib/bandit/pipeline.ex
+++ b/lib/bandit/pipeline.ex
@@ -47,6 +47,9 @@ defmodule Bandit.Pipeline do
         end
       rescue
         error -> handle_error(error, __STACKTRACE__, transport, span, opts)
+      catch
+        :throw, value ->
+          handle_throw(value, __STACKTRACE__, transport, span, opts)
       end
     rescue
       error ->
@@ -236,6 +239,17 @@ defmodule Bandit.Pipeline do
     else
       Bandit.HTTPTransport.send_on_error(transport, error)
       {:ok, transport}
+    end
+  end
+
+  defp handle_throw(value, stacktrace, transport, span, opts) do
+    Bandit.Telemetry.span_exception(span, :throw, value, stacktrace)
+    status = 500
+
+    if status in Keyword.get(opts.http, :log_exceptions_with_status_codes, 500..599) do
+      Logger.error(Exception.format(:throw, value, stacktrace), domain: [:bandit])
+      Bandit.HTTPTransport.send_on_error(transport, value)
+      {:error, value}
     end
   end
 end
--- a/lib/bandit/pipeline.ex
+++ b/lib/bandit/pipeline.ex
@@ -46,15 +46,15 @@ defmodule Bandit.Pipeline do
             {:upgrade, adapter.transport, protocol, opts}
         end
       rescue
-        error -> handle_error(error, __STACKTRACE__, transport, span, opts)
+        error -> handle_error(:error, error, __STACKTRACE__, transport, span, opts)
       catch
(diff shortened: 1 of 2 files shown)


lib/bandit/pipeline.ex
```diff
@@ -238,4 +241,15 @@ defmodule Bandit.Pipeline do
...
+  defp handle_throw(value, stacktrace, transport, span, opts) do
+    Bandit.Telemetry.span_exception(span, :throw, value, stacktrace)
+    status = 500
+
+    if status in Keyword.get(opts.http, :log_exceptions_with_status_codes, 500..599) do
+      Logger.error(Exception.format(:throw, value, stacktrace), domain: [:bandit])
+      Bandit.HTTPTransport.send_on_error(transport, value)
+      {:error, value}
+    end
+  end
```
grzuy, line 254: I guess one possibility might be to make `handle_error` more generalized and also handle this `throw` case.

Didn't want to purse that refactor yet for the sake of better code review-ability here. But let me know if you rather prefer that now :-)

mtrudel: I don't see any reason why not to handle these in a combined `handle_error` function which takes `:error` or `:throw` as a parameter!

lib/bandit/pipeline.ex
```diff
@@ -225,17 +229,20 @@ defmodule Bandit.Pipeline do
     {:error, error}
   end
 
-  defp handle_error(error, stacktrace, transport, span, opts) do
-    Bandit.Telemetry.span_exception(span, :exit, error, stacktrace)
-    status = error |> Plug.Exception.status() |> Plug.Conn.Status.code()
+  defp handle_error(kind, reason, stacktrace, transport, span, opts) do
+    Bandit.Telemetry.span_exception(span, telemetry_kind(kind), reason, stacktrace)
```
grzuy, line 233: Had to add this `telemetry_kind/1` function to keep existing behavior of emitting `:exit` kind on an exception.

Feels like this should be `:error` instead, and `:exit` would be if we later also emit telemetry event for non-normal exits by augmentig the `catch` to also...

```elixir
catch
  :exit, reason ->
```

lib/bandit/pipeline.ex
grzuy: Cool. Updated.

grzuy:
Hi,

Opened a new pull request (https://github.com/mtrudel/bandit/pull/398) that should supersede this one.

lib/bandit/pipeline.ex
grzuy: > Had to add this `telemetry_kind/1` function to keep existing behavior of emitting `:exit` kind on an exception.

Solved this in the newer PR, by introducing a new telemetry event and keeping the original one for backwards compatibility: https://github.com/mtrudel/bandit/pull/398/files#diff-68626306df11d3e10964941997c6cdafbf587a9ec7de087141840655a47a1855R231-R236

grzuy:
Kept digging more on this and I realized what I report in #410.
I thought it might be useful to report clearly as an issue before fixing.

PR #398, opened 2024-09-16
Attempt to improve plug call exception handling and logging

grzuy:
Supersedes #396.

- Telementry events
  - Also emit span for throw
  - Also emit span for exit

- Logger
  - Lets logger handlers know about the original crash reason when handling bandit errors
    - References:
      - Elixir Logger `crash_reason`: https://github.com/elixir-lang/elixir/blob/4bbddcc053498ca28ea5ebb5039de9b3a97ddf82/lib/logger/lib/logger.ex#L130-L135
      - Broadway passing `crash_reason`: https://github.com/dashbitco/broadway/blob/9eaf2140b6c1f36b25f990a7c74052bece26aadc/lib/broadway/acknowledger.ex#L104-L113
      - Plug.Cowboy passing `crash_reason`: https://github.com/elixir-plug/plug_cowboy/blob/746d848baca3436ca74fdab2750d14070b92002b/lib/plug/cowboy/translator.ex#L42-L46


lib/bandit/pipeline.ex
```diff
@@ -45,13 +45,14 @@ defmodule Bandit.Pipeline do
             Bandit.Telemetry.stop_span(span, adapter.metrics, %{conn: conn})
             {:upgrade, adapter.transport, protocol, opts}
         end
-      rescue
-        error -> handle_error(error, __STACKTRACE__, transport, span, opts)
```
mtrudel, line 49: I know it's redundant with the catch block below (`raise` being realized by try/catch internally), but I don't *think* it's part of official spec; can we leave this clause here to make it explicit? Alternately, if the language *is* authoritative on this, that's fine too.

lib/bandit/telemetry.ex
```diff
@@ -60,7 +60,27 @@ defmodule Bandit.Telemetry do
...
+      * `connection_telemetry_span_context`: The span context of the Thousand Island `:connection`
+        span which contains this request
+      * `conn`: The `Plug.Conn` representing this connection. Not present in cases where `error`
+        is also set and the nature of error is such that Bandit was unable to successfully build
+        the conn
+      * `kind`: The kind of unexpected condition, typically `:error`, `:throw` or `:exit`
+      * `reason`: The exception, throw value or exit reason which caused this unexpected termination
+      * `stacktrace`: The stacktrace of the location which caused this unexpected termination
+
+  * `[:bandit, :request, :exception]` (DEPRECATED)
```
mtrudel, line 83: I realize that `exception` is (and always has been) the wrong term here, but I've deliberately been using it to mirror the naming conventions of [:telemetry.span/3](https://hexdocs.pm/telemetry/telemetry.html#span/3).

mtrudel:
A couple of changes requested to keep things consistent and explicit; other than those changes the rest of this looks great!

lib/bandit/pipeline.ex
grzuy: Right.

Despite working like that doesn't seem to be actually documented in https://hexdocs.pm/elixir/try-catch-and-rescue.html that you can catch all (including `:error`) with just `catch`.

Re-instating the `rescue` clause just in case.

lib/bandit/telemetry.ex
grzuy: Gotcha.

I only introduced a new event name just to not break the shape of the actual one when supporting more `kind`s.

Updating back to only `[:bandit, :request, :exception]`, with the correct `kind` and `reason` field, and maybe keeping the `exception` field in case `is_exception(reason)`, so that this doesn't break peoples telemetry handlers listening for that field.

It will break anyone matching on `kind: :exit` and `exception: exception`. This will now be corrected to `kind: :error` and `exception: exception`.

lib/bandit/pipeline.ex
mtrudel: Good call. It sounds like it's an implementation detail then, so we're safer to be explicit in case they ever change it in the language.

lib/bandit/telemetry.ex
mtrudel: That's fine

grzuy:
Fixed format.

Can't reproduce locally the `protocol_test.exs` failure though.

grzuy:
@mtrudel Just realized that Bandit was actually re-raising error in 1.4 instead of just "logging or silence" with changes in 1.5 with https://github.com/mtrudel/bandit/pull/342/files#diff-68626306df11d3e10964941997c6cdafbf587a9ec7de087141840655a47a1855L65.

Curious about the reason not to continue re-raising? Similar to what plug cowboy does...
- https://github.com/elixir-plug/plug_cowboy/blob/ced253f035f2a07ebed17cfacb64d4f16357a750/lib/plug/cowboy/handler.ex#L21-L23
- https://github.com/elixir-plug/plug_cowboy/blob/ced253f035f2a07ebed17cfacb64d4f16357a750/lib/plug/cowboy/handler.ex#L52-L63

In case there's the possibility of introducing re-raise again, this pull request could be re-worked.

grzuy:
Reading this thread also got me thinking about it https://groups.google.com/g/elixir-lang-core/c/pWz-uTVMEVM/m/QGXncxD1AQAJ

grzuy:
Mmm... not sure to he honest.

I'm exploring the "re-raising approach" in a separate branch to compare.

I'll let you know how it goes.

grzuy:
> I'm exploring the "re-raising approach" in a separate branch to compare.

Opening draft with what I got from playing with that alternative approach:  https://github.com/mtrudel/bandit/pull/400.

lib/bandit/pipeline.ex
```diff
@@ -46,12 +46,15 @@ defmodule Bandit.Pipeline do
             {:upgrade, adapter.transport, protocol, opts}
         end
       rescue
-        error -> handle_error(error, __STACKTRACE__, transport, span, opts)
+        error -> handle_error(:error, error, __STACKTRACE__, transport, span, opts)
+      catch
+        kind, reason ->
+          handle_error(kind, reason, __STACKTRACE__, transport, span, opts)
```
grzuy, line 52: This is not right.
It will catch and handle "normal" exists as errors.

grzuy:
Closing in favor smaller and more focused #411 for now.
Will revisit and build on top of #411 if that one accepted and merged.

PR #400, opened 2024-09-18
Error handling and logging improvements via re-raising

grzuy:
Potential alternative to #398 .

Proof of concept still.

Haven't yet tested how compatible is with earlier elixir/otp versions.

This moves to do something "similar" to what plug_cowboy does in https://github.com/elixir-plug/plug_cowboy/blob/ced253f035f2a07ebed17cfacb64d4f16357a750/lib/plug/cowboy/handler.ex#L21-L23.

### DIFF
--- a/lib/bandit/pipeline.ex
+++ b/lib/bandit/pipeline.ex
@@ -227,6 +227,7 @@ defmodule Bandit.Pipeline do
   end
 
   defp handle_error(kind, reason, stacktrace, transport, span, opts) do
+    # TODO: Fix this telementry kind
     Bandit.Telemetry.span_exception(span, :exit, reason, stacktrace)
     status = reason |> Plug.Exception.status() |> Plug.Conn.Status.code()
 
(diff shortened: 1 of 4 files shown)


lib/bandit/pipeline.ex
```diff
@@ -216,16 +226,17 @@ defmodule Bandit.Pipeline do
...
 
-  defp handle_error(error, stacktrace, transport, span, opts) do
-    Bandit.Telemetry.span_exception(span, :exit, error, stacktrace)
-    status = error |> Plug.Exception.status() |> Plug.Conn.Status.code()
+  defp handle_error(kind, reason, stacktrace, transport, span, opts) do
+    # TODO: Fix this telementry kind
+    Bandit.Telemetry.span_exception(span, :exit, reason, stacktrace)
+    status = reason |> Plug.Exception.status() |> Plug.Conn.Status.code()
 
     if status in Keyword.get(opts.http, :log_exceptions_with_status_codes, 500..599) do
```
grzuy, line 234: Ideally this conditional would be inside the `Bandit.LoggerTranslator`. Need a way to access the `opts.http` from there.

grzuy:
Not sure this will land at something worth merging.
Closing for now as stale...

PR #405, opened 2024-10-10
Fix typo

josevalim:


mtrudel:
And here I was all chuffed with my new spellcheck setup in neovim that corrects comments (but not atom names). Thanks for the PR!

PR #411, opened 2024-10-23
fix: throwing plug properly handled and returns 500

### LINKED ISSUE #410: Mishandled plug `throw` value
The value thrown by a plug, because is not handled by bandit, ends up being treated as the 2nd element value in an `{:ok, value}` result of `ThousandIsland.Handler.handle_connection`.

So, when I test in a plug `throw "something"`, I end up getting a `ThousandIsland.Handler` abnormal termination of reason `{:bad_return_value, "something"}`.

Digging into the call stack I realized this means that erlang's `gen_server` catches the thrown value and tries to interpret it as the value to be return by the `gen_server` callback, in this case `ThousandInsland.Handler.handle_continue`.

- gen_server [loop](https://github.com/erlang/otp/blob/44ffe8811dfcf3d2fe04d530c6e8fac5ca384e02/lib/stdlib/src/gen_server.erl#L2244-L2248)
  - [try_handle_continue](https://github.com/erlang/otp/blob/44ffe8811dfcf3d2fe04d530c6e8fac5ca384e02/lib/stdlib/src/gen_server.erl#L2333-L2341)
  - [handle_common_reply](https://github.com/erlang/otp/blob/44ffe8811dfcf3d2fe04d530c6e8fac5ca384e02/lib/stdlib/src/gen_server.erl#L2479-L2480)

This means that I was getting `:bad_return_value` just because `"something"` is an invalid value for `gen_server` callback returns. But if I `throw {:noreply, {}}` in the plug, the `gen_server` doesn't terminate, and it's state is updated to `{}` or whatever other value is returned, right?

---

For what is worth, this is the output I get when placing a few inspect statement along `bandit` and `thousand_island` while testing and debugging.

```
"======== ThousandIsland.Handler.init"
"======== ThousandIsland.Handler.handle_info"
"======== ThousandIsland.Handler.handle_continue"
"======== Bandit.DelegatingHandler.handle_connection"
"======== Bandit.DelegatingHandler.handle_bandit_continuation"
"======== Bandit.DelegatingHandler.handle_connection"
"======== Bandit.DelegatingHandler.handle_bandit_continuation"
"======== Bandit.DelegatingHandler.handle_data"
"======== Bandit.HTTP1.Handler.handle_data - calling Pipeline.run"
"======== ThousandIsland.Handler.terminate - reason: {:bad_return_value, \"something\"}"
```

grzuy:
closes #410.

An re-worked attempt after #396, trying to improve it by taking some pieces from #398.

### DIFF
--- a/lib/bandit/telemetry.ex
+++ b/lib/bandit/telemetry.ex
@@ -76,8 +76,8 @@ defmodule Bandit.Telemetry do
       * `conn`: The `Plug.Conn` representing this connection. Not present in cases where `error`
         is also set and the nature of error is such that Bandit was unable to successfully build
         the conn
-      * `kind`: The kind of unexpected condition, typically `:exit`
-      * `exception`: The exception which caused this unexpected termination
+      * `kind`: The kind of unexpected condition, typically `:error` or `:throw`
+      * `reason`: The error reason. An Exception struct or the thrown value
       * `stacktrace`: The stacktrace of the location which caused this unexpected termination
 
   ## `[:bandit, :websocket, *]`
@@ -188,14 +188,22 @@ defmodule Bandit.Telemetry do
     untimed_span_event(span, :stop, measurements, metadata)
   end
 
-  @spec span_exception(t(), Exception.kind(), Exception.t(), Exception.stacktrace()) :: :ok
-  def span_exception(span, kind, exception, stacktrace) do
+  @spec span_exception(t(), Exception.kind(), Exception.t() | term(), Exception.stacktrace()) :: :ok
+  def span_exception(span, kind, reason, stacktrace) do
     metadata =
       Map.merge(span.start_metadata, %{
         kind: kind,
-        exception: exception,
+        reason: reason,
         stacktrace: stacktrace
       })
+
+    # Keeps exception field for backwards compatibility
+    metadata =
+      if kind == :error and is_exception(reason) do
+        Map.put(metadata, :exception, reason)
+      else
+        metadata
+      end
 
     span_event(span, :exception, %{}, metadata)
   end
--- a/lib/bandit/telemetry.ex
+++ b/lib/bandit/telemetry.ex
@@ -188,7 +188,8 @@ defmodule Bandit.Telemetry do
     untimed_span_event(span, :stop, measurements, metadata)
   end
 
-  @spec span_exception(t(), Exception.kind(), Exception.t() | term(), Exception.stacktrace()) :: :ok
+  @spec span_exception(t(), Exception.kind(), Exception.t() | term(), Exception.stacktrace()) ::
+          :ok
   def span_exception(span, :error, reason, stacktrace) when is_exception(reason) do
     metadata =
       Map.merge(span.start_metadata, %{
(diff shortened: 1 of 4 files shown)


lib/bandit/telemetry.ex
```diff
@@ -76,8 +76,8 @@ defmodule Bandit.Telemetry do
       * `conn`: The `Plug.Conn` representing this connection. Not present in cases where `error`
         is also set and the nature of error is such that Bandit was unable to successfully build
         the conn
-      * `kind`: The kind of unexpected condition, typically `:exit`
-      * `exception`: The exception which caused this unexpected termination
+      * `kind`: The kind of unexpected condition, typically `:error` or `:throw`
+      * `reason`: The error reason. An Exception struct or the thrown value
```
mtrudel, line 80: The rest of this PR looks great; these are the only two items I have issue with.

For better or worse, we've already baked in `exception` and `exit` in metadata, and [downstream packages](https://github.com/open-telemetry/opentelemetry-erlang-contrib/blob/main/instrumentation/opentelemetry_bandit/lib/opentelemetry_bandit.ex) already depend on them. We can't change them without a major version bump and that's not happening for this. 

I'd suggest:

* Changing `:error` back to `:exit` and `reason` back to `exception`
* Keeping `:throw` as a second valid value for`kind`
* Adding a new `:throw_value` key to pass the thrown value if it's a throw

This is enough to be backwards compatible and still have sensible semantic here I think.

mtrudel (changes requested):

grzuy:
Conflicts with `main` expected after #413 merged.
Solving soon.

lib/bandit/telemetry.ex
grzuy: Thanks for the feedback!

Sounds fair.

Related to this previous discussion, for what is worth, https://github.com/mtrudel/bandit/pull/398#discussion_r1761940338.

Will try adapt to something backwards-compatible.

mtrudel:
let me know when this is ready to stand for review again!

lib/bandit/telemetry.ex
grzuy: After digging into `opentelemetry_bandit` I realized it might be a problem to start emitting events of `kind: :throw` right now. `opentelemetry_bandit` (actually `opentelemetry_api`) is expecting only exceptions: https://github.com/open-telemetry/opentelemetry-erlang/blob/main/apps/opentelemetry_api/lib/open_telemetry/span.ex.

The story is different for cowboy because it's using `otel_span.erl`, which actually handles gracefully different kinds (referred to as `Class` in there): https://github.com/open-telemetry/opentelemetry-erlang/blob/a506d08e8439c353bbf8277a1ca619c668d125f8/apps/opentelemetry_api/src/otel_span.erl#L239-L242. 

Rolling back for now, to only emitting the same events as before this PR and later can tackle emitting new throw events in tandem with possible updates in `opentelemetry_bandit`.

grzuy: > Rolling back for now, to only emitting the same events as before this PR and later can tackle emitting new throw events in tandem with possible updates in `opentelemetry_bandit`.

https://github.com/mtrudel/bandit/pull/411/commits/e8239b612a8003f05effe657e862aa37b451b412

grzuy:
Thank you :pray:

mtrudel:
Thank YOU! Sorry this took so long to get sorted; i've had my hands full with real world things and am finally getting back to being able to spend some time on Bandit. Thanks for the patience.

grzuy:
No worries!
Thanks for all your great work :100:

PR #420, opened 2024-11-13
feat: allow logger handler to know about HTTP request details

grzuy:
Following up #417.

Another piece that helps switch from `plug_cowboy` to `bandit` without losing features in this Error Tracking / Logging area.

In `plug_cowboy`: https://github.com/elixir-plug/plug_cowboy/pull/57.

Example of a logger handler that consumes the `conn` made available as metadata by plug_cowboy:
- `logger_json`: https://hexdocs.pm/logger_json/LoggerJSON.html#module-metadata.

### DIFF
--- a/lib/bandit/pipeline.ex
+++ b/lib/bandit/pipeline.ex
@@ -46,10 +46,10 @@ defmodule Bandit.Pipeline do
             {:upgrade, adapter.transport, protocol, opts}
         end
       rescue
-        exception -> handle_error(:error, exception, __STACKTRACE__, transport, span, opts)
+        exception -> handle_error(:error, exception, __STACKTRACE__, transport, span, opts, conn)
       catch
         :throw, value ->
-          handle_error(:throw, value, __STACKTRACE__, transport, span, opts)
+          handle_error(:throw, value, __STACKTRACE__, transport, span, opts, conn)
       end
     rescue
       exception ->
@@ -202,9 +202,12 @@ defmodule Bandit.Pipeline do
           Exception.stacktrace(),
           Bandit.HTTPTransport.t(),
           Bandit.Telemetry.t(),
-          map()
+          map(),
+          Plug.Conn.t() | nil
         ) :: {:ok, Bandit.HTTPTransport.t()} | {:error, term()}
-  defp handle_error(:error, %type{} = error, stacktrace, transport, span, opts)
+  defp handle_error(kind, error, stacktrace, transport, span, opts, conn \\ nil)
+
+  defp handle_error(:error, %type{} = error, stacktrace, transport, span, opts, _conn)
        when type in [
               Bandit.HTTPError,
               Bandit.HTTP2.Errors.StreamError,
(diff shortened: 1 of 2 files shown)


mtrudel:
Correct me if i'm wrong, but this is pushing the entirety of the conn across to the logger and is thus ending up in people's logging stacks correct? If so, this is a bit of a deal breaker (at least by default) since it would be also exposing cookies and other sensitive bits as part of that.

grzuy:
> Correct me if i'm wrong, but this is pushing the entirety of the conn across to the logger and is thus ending up in people's logging stacks correct?

It depends, but probably not.

It will depend on what logger formatter the application is using and how it is configured.

That said, I don't expect any formatters to blindly push all metadata blindly. Given that 

Elixir `Logger`'s default formatter will not print any metadata unless the key is explicitly listed in the configuration. So in that case, will not be included in printed logs by default.

For Elixir applications sending logs to third-party services, I guess it depends on what their built-in logger formatters do by default. I may be wrong, but I expect none to, by default, include all logger metadata blindly.

To go over one example, `logger_json`, which supports sending logs as JSON to several popular services, will, by default, NOT send any metadata as far as I understand, per

image
image

source: https://hexdocs.pm/logger_json/6.2.1/LoggerJSON.html#module-shared-options

It can be set to `metadata: {:all_except, [:conn]}`, which is given as an example in its README: https://github.com/Nebo15/logger_json?tab=readme-ov-file#configuration.

Then, each of the 4 built-in formatters use that `:conn` metadata key, to extract useful request details before reporting to these 3rd party following their expected JSON schema.

- [`LoggerJSON.Formatters.Basic.format_http_request/1`](https://github.com/Nebo15/logger_json/blob/584a6243b66dbd074c1e87a177a278070005ab12/lib/logger_json/formatters/basic.ex#L76-L95)
- [`LoggerJSON.Formatters.Datadog.format_http_request/1`](https://github.com/Nebo15/logger_json/blob/584a6243b66dbd074c1e87a177a278070005ab12/lib/logger_json/formatters/datadog.ex#L199-L230)
- [`LoggerJSON.Formatters.Elastic.format_http_request/1`](https://github.com/Nebo15/logger_json/blob/584a6243b66dbd074c1e87a177a278070005ab12/lib/logger_json/formatters/elastic.ex#L275-L296)
- [`LoggerJSON.Formatters.GoogleCloud.format_http_request/1`](https://github.com/Nebo15/logger_json/blob/584a6243b66dbd074c1e87a177a278070005ab12/lib/logger_json/formatters/google_cloud.ex#L243-L266)


Worth noting that the presence of the `conn` in the logger metadata by default is something that plug_cowboy has since 2020.

grzuy:
Thanks Mat!

For what is worth, been running a phoenix production app against bandit `main` since yesterday, no issues.

Also tested:

- A throw was handled :heavy_check_mark: 
- `crash_reason` and `conn` are sent as logger metadata :heavy_check_mark:

PR #421, opened 2024-11-15
Deflate support on chunked responses

mtrudel:
Adds support for deflating chunked responses. Consists of a refactor to first break apart the existing compression routines such that the constituent stages can be called separately throughout the lifecycle of a chunked response, and then using these new routines within chunked responses.

Only deflate is supported on responses


mtrudel:
Backed this out of main/1.6.0 since there are reports of it breaking. Will re-open and fix

mtrudel:
Reopened as #429

Issue #392, opened 2024-08-20
truncated responses with H2C (via AWS Application Load Balancer)


paulswartz:
I recently tried running a Bandit-based Phoenix application with H2C, behind an AWS Application Load Balancer. In this mode, I would almost never receive a full response, even for static files served by `Plug.Static`.

Some alternatives which did work, making me thing that the issue is with Bandit's H2C implementation:
- Bandit via HTTP/1.1
- Cowboy 2.12 via H2C

I don't have much in the way of logs at the moment, but if there's anything that would be helpful, I'm happy to give it a shot!

mtrudel:
I'd love a repro case here, not of the application itself, but of the configuration it took within AWS to get here! I'm willing to bet it rhymes with #306 and/or #338; likely a bug in who / how content-length headers get set.

paulswartz:
For us, it's:
- Phoenix 1.7 app
- Debian bookworm Docker container
- AWS ECS Fargate runner
- AWS Application Load Balancer
- AWS Target group configured with the `HTTP2` [target version](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-target-groups.html#target-group-protocol-version) to talk to the ECS Fargate tasks

Good response:
![Screenshot 2024-08-20 at 18 50 35](https://github.com/user-attachments/assets/96117a7a-dfcf-4cf1-b11c-a2334772d692)

Bad response:
![Screenshot 2024-08-20 at 18 50 45](https://github.com/user-attachments/assets/5c69035e-007c-4c5f-9856-1c23cc348e96)

Server logs:
```
023cd54e6718 22:47:13.750 [info] state=sent method=GET path=/assets/app-b49524acc1698e2440c53a73d41aec25.js params=[UNFETCHED] status=200 headers={"user-agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:129.0) Gecko/20100101 Firefox/129.0"} duration=0.248
023cd54e6718 22:47:14.649 [info] state=sent method=GET path=/assets/app-b49524acc1698e2440c53a73d41aec25.js params=[UNFETCHED] status=200 headers={"user-agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:129.0) Gecko/20100101 Firefox/129.0"} duration=0.234
```

paulswartz:
From cURL:
```
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
[...]

PR #427, opened 2024-11-18
Remove hardcoded headers

lucacorti:
I've tested the fix for #392 and it works for me, after removing the hardcoded response headers.


mtrudel:
Thanks for this but there's no need - the aws_experiments branch was just for testing; i'm not planning on landing any shipping code there.

PR #428, opened 2024-11-18
Use inform to send websocket upgrades instead of a regular sending

### LINKED ISSUE #422: WSS broken by frame masking
image

Maybe this is just happening for me. I am investigating.

mtrudel:
Fixes #422


mtrudel:
I honestly don't know what I was thinking when I originally wrote this with send_resp instead of inform

PR #429, opened 2024-11-18
Deflate support on chunked responses (take 2)

mtrudel:
Re-opening of #421, which GitHub got confused about after I forced pushed it away


mtrudel:
Ready for review, with a few improvements for 204, 304, and empty body responses

PR #433, opened 2024-11-29
Improve unrecoverable error handling

### LINKED ISSUE #432: Bandit returns :closed error as a string for chunk
I got a CSV export controller that in production occassionally gives a match error, because I'm expecting `{:error, :closed}` for `Plug.Conn.chunk/2` as the example from [the docs](https://hexdocs.pm/plug/Plug.Conn.html#chunk/2) shows:

```elixir
Enum.reduce_while(~w(each chunk as a word), conn, fn (chunk, conn) ->
  case Plug.Conn.chunk(conn, chunk) do
    {:ok, conn} ->
      {:cont, conn}
    {:error, :closed} ->
      {:halt, conn}
  end
end)
```

However, Bandit returns a string (`{:error, "closed"}`) on this line:
https://github.com/mtrudel/bandit/blob/84e0d5d85c122724056607e60fc37164ba9e4347/lib/bandit/adapter.ex#L204

Which ultimately is from this line for the HTTP/1 socket: https://github.com/mtrudel/bandit/blob/84e0d5d85c122724056607e60fc37164ba9e4347/lib/bandit/http1/socket.ex#L428-L430

When I tested with HTTP/2 it seems like the process just continues to completion even though I closed the connection client side? If I let it wait for a long time it just silently stops.

So I've got two questions:

1. Shouldn't the HTTP/1 adapter return the `:closed` atom? I think we've looked at this before with no specified return types, but the example does show that `Plug.Conn.chunk/2` returns `{:error, :closed}` (so at least the example needs to be updated there to show both cowboy and bandit).

2. Is it correct behavior to let HTTP/2 continue to process the request even though the client cancelled the request? Maybe I was testing this wrong or I don't understand how the HTTP/2 stream is supposed to work, but I couldn't get it to stop processing the request though I cancelled the download and closed the connection in the browser.

mtrudel:
Fixes #432

### DIFF
--- a/test/bandit/http1/request_test.exs
+++ b/test/bandit/http1/request_test.exs
@@ -2448,7 +2448,7 @@ defmodule HTTP1RequestTest do
       conn = send_chunked(conn, 200)
 
       Process.sleep(200)
-      assert chunk(conn, "IMPOSSIBLE") == {:error, "closed"}
+      assert chunk(conn, "IMPOSSIBLE") == {:error, :closed}
 
       conn
     end
--- a/test/bandit/http1/request_test.exs
+++ b/test/bandit/http1/request_test.exs
@@ -111,7 +111,7 @@ defmodule HTTP1RequestTest do
           Process.sleep(500)
         end)
 
-      assert output =~ "[error] ** (Bandit.HTTPError) closed"
+      assert output =~ "[error] ** (Bandit.SocketError) Unrecoverable socket error: closed"
 
       # Make sure we don't log a stacktrace
       refute output =~ "lib/bandit/pipeline.ex:"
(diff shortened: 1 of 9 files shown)


test/bandit/http1/request_test.exs
```diff
@@ -2398,7 +2398,7 @@ defmodule HTTP1RequestTest do
           Process.sleep(500)
         end)
 
-      assert output =~ "(Bandit.HTTPError) closed"
+      assert output =~ "(Bandit.TransportError) Unrecoverable error: closed"
```
robredmond, line 2401: Add "[error] ** " prefix for consistency?

robredmond:
> @robredmond updated!

Haha, you're too quick!  I reread, realized it was a regex match and the prefix didn't really matter and deleted my comment

Issue #431, opened 2024-11-25
Continuation for problems under high load


atavistock:
Followup from https://github.com/mtrudel/bandit/issues/412

`ab` is definitely now passing, but unfortunately the benchmarks are not.  ;(

They're actually using `wrk` to do the benchmarking.  I was able to reproduce the issue easily with `ab` but its still breaking under load.  After seeing it pass with `ab` and not pass in the benchmarks I reproduced the problem with `wrk`.

To reproduce:
- Make `pipeline.lua` with:
```
init = function(args)
  local r = {}
  local depth = tonumber(args[1]) or 1
  for i=1,depth do
    r[i] = wrk.format()
  end
  req = table.concat(r)
end

request = function()
  return req
end
```
- run
`wrk -H 'Host: localhost' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 1024 --timeout 8 -t 10 http://localhost:4000/plaintext -s ./pipeline.lua -- 16`

**edited Nov 27, 2024**

There is a flood of `[error] ** (Bandit.HTTPError) Request line HTTP error: "\n"` before the server stops responding.  

For one, I'm not really sure what that error means.  It got an empty line as a request?

Two, is it possible that when these occur its not closing the connection?

atavistock:
I should mention I just threw this into an existing toy project.  Let me make sure its happening in a clean env.

robredmond:
@atavistock - curious if you were seeing either of the following as well?
* `[error] ** (Bandit.HTTPError) Header read timeout`
* `[error] ** (Bandit.HTTPError) Excess body read`

atavistock:
Not that I've seen.

It seems to be mostly `** (Bandit.HTTPError) Request line HTTP error: "\n"`

atavistock:
Not that I've seen.

It seems to be mostly `[error] ** (Bandit.HTTPError) Request line HTTP error: "\n"`

atavistock:
Theres also a flood of 
```
Handler :default switched from :sync to :drop mode
Handler :default switched from :drop to :sync mode
```
before the other errors

mtrudel:
> [notice] Handler :default switched from :sync to :drop mode
[notice] Handler :default switched from :drop to :sync mode

Those definitely aren't coming from Bandit; likely they are logger as you suspect. 

[...]

PR #437, opened 2024-12-04
Handle pipelined requests

mtrudel:
Fixes the underlying issue in #431.

### DIFF
--- a/lib/bandit/http1/handler.ex
+++ b/lib/bandit/http1/handler.ex
@@ -25,7 +25,16 @@ defmodule Bandit.HTTP1.Handler do
       if Keyword.get(state.opts.http_1, :clear_process_dict, true), do: clear_process_dict()
       gc_every_n_requests = Keyword.get(state.opts.http_1, :gc_every_n_keepalive_requests, 5)
       if rem(requests_processed, gc_every_n_requests) == 0, do: :erlang.garbage_collect()
-      {:continue, Map.put(state, :requests_processed, requests_processed)}
+
+      state = Map.put(state, :requests_processed, requests_processed)
+
+      # We have bytes that we've read but haven't yet processed, tail call handle_data to start
+      # reading the next request
+      if IO.iodata_length(transport.buffer) != 0 do
+        handle_data(transport.buffer, transport.socket, state)
+      else
+        {:continue, state}
+      end
     else
       {:close, state}
     end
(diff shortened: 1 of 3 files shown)


lib/bandit/http1/handler.ex
```diff
@@ -25,7 +25,16 @@ defmodule Bandit.HTTP1.Handler do
       if Keyword.get(state.opts.http_1, :clear_process_dict, true), do: clear_process_dict()
       gc_every_n_requests = Keyword.get(state.opts.http_1, :gc_every_n_keepalive_requests, 5)
       if rem(requests_processed, gc_every_n_requests) == 0, do: :erlang.garbage_collect()
-      {:continue, Map.put(state, :requests_processed, requests_processed)}
+
+      state = Map.put(state, :requests_processed, requests_processed)
+
+      # We have bytes that we've read but haven't yet processed, tail call handle_data to start
+      # reading the next request
+      if IO.iodata_length(transport.buffer) != 0 do
```
xinz, line 33:
```suggestion
      if transport.buffer != <<>> do
```

One small suggestion with the following compare `bytes != <<>>` and `IO.iodata_length(bytes) != 0`:

```
Name                                             ips        average  deviation         median         99th %
bytes != <<>> with large bytes              216.30 M        4.62 ns  ±6611.15%           0 ns          11 ns
bytes != <<>> with empty bytes              199.46 M        5.01 ns  ±6209.18%           0 ns          11 ns
bytes != <<>> with small bytes              189.55 M        5.28 ns  ±6057.20%           0 ns          13 ns
IO.iodata_length != 0 with large bytes       39.31 M       25.44 ns ±34060.36%          15 ns          41 ns
IO.iodata_length != 0 with samll bytes       37.95 M       26.35 ns ±39355.33%          16 ns          44 ns
IO.iodata_length != 0 with empty bytes       35.85 M       27.89 ns ±33175.52%          16 ns          45 ns

Comparison:
bytes != <<>> with large bytes              216.30 M
bytes != <<>> with empty bytes              199.46 M - 1.08x slower +0.39 ns
bytes != <<>> with small bytes              189.55 M - 1.14x slower +0.65 ns
IO.iodata_length != 0 with large bytes       39.31 M - 5.50x slower +20.82 ns
IO.iodata_length != 0 with samll bytes       37.95 M - 5.70x slower +21.73 ns
IO.iodata_length != 0 with empty bytes       35.85 M - 6.03x slower +23.27 ns
```

mtrudel: I'm deliberately not doing this because 1. It's almost always the case that this is empty (it's not not if the client is pipelining, which is pretty rare), and 2. since it's an iolist there's uncountably many ways that it could be empty (`<<>>`, `[]`, [<<>>], `[<<>>, []]`, etc). So there's no performance hit in the most common case of it being empty, and it's always correct.

xinz: Thanks very much for clarification.

PR #443, opened 2024-12-27
Add support for sending telemetry events on throws and exits

mtrudel:
Relevant to your interests @martasaur. 

@bryannaegele, is this compatible downstream as-is? Any changes I can / should make to help if not?

### DIFF
--- a/lib/bandit/telemetry.ex
+++ b/lib/bandit/telemetry.ex
@@ -80,7 +80,8 @@ defmodule Bandit.Telemetry do
         the conn
       * `plug`: The Plug which is being used to serve this request. Specified as `{plug_module, plug_opts}`
       * `kind`: The kind of unexpected condition, typically `:exit`
-      * `exception`: The exception which caused this unexpected termination
+      * `exception`: The exception which caused this unexpected termination. May be an exception
+        or an arbitrary value when the event was an uncaught throw or an exit
       * `stacktrace`: The stacktrace of the location which caused this unexpected termination
 
   ## `[:bandit, :websocket, *]`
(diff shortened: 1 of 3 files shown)


bryannaegele:
Looks like it's just expanding the scope of the current functionality and the signature is the same, so I think it would "just work"?

mtrudel:
Yep, strictly additive

grzuy:
Won't it break in

https://github.com/open-telemetry/opentelemetry-erlang/blob/25d755e/apps/opentelemetry_api/lib/open_telemetry/span.ex#L138

?

It seems `Span.record_exception` expects an `Exception` struct. Other terms won't work apparently.

FWIW https://github.com/open-telemetry/opentelemetry-erlang/discussions/798.

grzuy:
Oh, just saw the `when is_exception?` guard.
Nevermind.
It seems it will be a no-op.

Sorry for the noise.

grzuy:
It seems there's still an issue with this...

The call to `Tracer.record_exception(meta.exception, meta.stacktrace)` is a no-op as I said before, but a few lines after that it crashes the telemetry handler when evaluating `meta.exception.__struct`, when the exception is a throw or exit value, instead of an exception struct.

https://github.com/open-telemetry/opentelemetry-erlang-contrib/blob/b6104221a0f40dfce61afe2a2a1798039fd4ffee/instrumentation/opentelemetry_bandit/lib/opentelemetry_bandit.ex#L517.

grzuy:
Drafting a solution in https://github.com/open-telemetry/opentelemetry-erlang-contrib/pull/467.

PR #446, opened 2025-01-02
Improve tests

mtrudel:
* Run tests async
* Get rid of `sleep` calls in tests that were there only to (racily!) capture logs / telemetry, and replace with `receive` based log / telemetry test helpers
* Test suite speedup of about **10x!**
* Add `plug` to telemetry metadata
* Add `plug` to logging metadata
* Improve many sad path logging cases to include a full suite of metadata


mtrudel:
In the interest of review ability I'm going to merge this even though there's still some test improvements to be done. Things 'coming soon' after this goes out:

* Use the same message based approach to log testing for WebSockets. Requires a rethink of how we do logging in the WebSocket stack
* Split `http1/request_test.exs` into `http1/{plug, protocol}_test.exs` to mirror the pattern for http2 and websocket

PR #452, opened 2025-01-07
Always close HTTP/1 connection in any case where an error comes out of the plug

### LINKED ISSUE #451: Wrong error handling when receiving invalid JSON within a keepalive connection
Thank you for providing the package! 

I'm unsure if this is a bug in Bandit or Plug. The side effect is interesting. In an environment with keep-alive connections, sending a request with invalid JSON leads to the cancellation of the subsequent request. 

The first request ends with an exception in the plug parser. The state of the buffer that has already been read is dropped and the payload of the subsequent request is appended to this buffer. The result is that fragments of the invalid JSON are now located before the new request. Bandit cannot parse the request header and therefore the connection is closed:

```
[error] ** (Bandit.HTTPError) Request line HTTP error: "{\"hello\": broken}POST /api HTTP/1.1\r\n"
```

I published a git repository to reproduce this issue: https://github.com/zookzook/bandit-plug-parser-test

In my opinion, the error lies in the error handling by exceptions, which causes the state of the buffer to be lost. Conversely, one could argue that Bandit should terminate the connection with the 400 status code.  Or the parser should use a 500 status code instead of a 400. For my part, I solved the problem with my own parser that uses a 500 status code. 

I have no particular opinion on this and leave it to the authors (i.e., you) to do the right thing :-)

Either way, this issue also serves as a reference for this situation and how to solve it. Of course, you can ask yourself why the client is sending invalid JSON in the first place, but I have discussed this often and at length with the frontend developers. In the end, the server must ensure that the error handling does not affect subsequent requests.

Background:

In our environment, we use Traefik as an ingress controller, which uses the keepalive connections. We monitor from time to time 520er errors as a result of the Cloudflare proxy. The frontend-client uses a retry mechanism in the case of an 400 status code. So, the first request ends in a 400 status code. After that, the client tries the same requests for 8 times before giving up (but the client will start again after a while). That means, 1+8 following random requests ended in a 520er error.

mtrudel:
Fixes #451 

Previously, Bandit would keep the connection open for errors which it did not log; this was an oversight.

### DIFF
--- a/lib/bandit/pipeline.ex
+++ b/lib/bandit/pipeline.ex
@@ -236,12 +236,9 @@ defmodule Bandit.Pipeline do
     if status in Keyword.get(opts.http, :log_exceptions_with_status_codes, 500..599) do
       logger_metadata = Bandit.Logger.logger_metadata_for(kind, reason, stacktrace, metadata)
       Logger.error(Exception.format(kind, reason, stacktrace), logger_metadata)
-
-      Bandit.HTTPTransport.send_on_error(transport, reason)
-      {:error, reason}
-    else
-      Bandit.HTTPTransport.send_on_error(transport, reason)
-      {:ok, transport}
     end
+
+    Bandit.HTTPTransport.send_on_error(transport, reason)
+    {:error, reason}
   end
 end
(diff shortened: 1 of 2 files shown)


zookzook (approved):
Looks good to me! Fast PR!

mtrudel:
Published as 1.6.3

PR #455, opened 2025-01-13
fix: properly normalize erlang errors before emitting telemetry and logged crash_reason

grzuy:
Hi there :wave: 

Minor follow up for https://github.com/mtrudel/bandit/commit/a053c1cd95d90756bf1dd6927199f8f2afc5b9fb.

Analogous code in `plug_cowboy`: https://github.com/elixir-plug/plug_cowboy/blob/ced253f035f2a07ebed17cfacb64d4f16357a750/lib/plug/cowboy/handler.ex#L51-L55.

Thanks!

### DIFF
--- a/test/bandit/http1/plug_test.exs
+++ b/test/bandit/http1/plug_test.exs
@@ -112,7 +112,7 @@ defmodule HTTP1PlugTest do
       get "/" do
         # Quiet the compiler
         _ = conn
-        raise "boom"
+        1 = 0
       end
     end
 
@@ -134,9 +134,15 @@ defmodule HTTP1PlugTest do
       SimpleHTTP1Client.send(client, "GET", "/hello_world", ["host: banana"])
       assert SimpleHTTP1Client.connection_closed_for_reading?(client)
 
-      assert_receive {:log, %{level: :error, msg: {:string, msg}}}, 500
+      assert_receive {:log, %{level: :error, msg: {:string, msg}, meta: meta}}, 500
       refute msg =~ "(Plug.Conn.WrapperError)"
-      assert msg =~ "** (RuntimeError) boom"
+      assert msg =~ "** (MatchError)"
+
+      assert %{
+               domain: [:elixir, :bandit],
+               crash_reason: {%MatchError{}, [_ | _] = _stacktrace},
+               conn: %Plug.Conn{}
+             } = meta
     end
   end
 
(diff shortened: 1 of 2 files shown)


test/bandit/http1/plug_test.exs
```diff
@@ -134,9 +134,15 @@ defmodule HTTP1PlugTest do
       SimpleHTTP1Client.send(client, "GET", "/hello_world", ["host: banana"])
       assert SimpleHTTP1Client.connection_closed_for_reading?(client)
 
-      assert_receive {:log, %{level: :error, msg: {:string, msg}}}, 500
+      assert_receive {:log, %{level: :error, msg: {:string, msg}, meta: meta}}, 500
       refute msg =~ "(Plug.Conn.WrapperError)"
-      assert msg =~ "** (RuntimeError) boom"
+      assert msg =~ "** (MatchError)"
```
mtrudel, line 139: What's the reason for this change? Is there an issue that comes up with MatchError that doesn't with a simple raise?

mtrudel: I see. I wonder if we should be doing that for all errors (ie continue to have the `Plug.Conn.WrapperError` pattern match be 'dumb' and just pass the value through unchanged) and normalize in the other handle_error clauses. Does that make sense, or is it only needed in the narrow case of wrapped errors?

grzuy: Yes, I think you're right.
That's regardless of whether it's wrapped or not.

grzuy: Updated.

grzuy: Also now realized this will fix/normalize the emitted telemetry exception.

Updated PR title.

mtrudel:
Looks great! Thanks for the continued work!

mtrudel:
@grzuy I just noticed that this PR introduced console warnings during tests; would it be sufficient for us to explicitly raise a `MatchError` / `ArithmeticError` in the relevant places?

grzuy:
> @grzuy I just noticed that this PR introduced console warnings during tests;

Hey @mtrudel, sorry about that.

https://github.com/mtrudel/bandit/pull/459 should fix one.
Will also fix the other warning soon-ish.

> would it be sufficient for us to explicitly raise a `MatchError` / `ArithmeticError` in the relevant places?

No, because the test cases are there to test that when original erlang `:badarith` or `{:badmatch, x}` are generate, Bandit properly normalizes. If we directly raise the Elixir normalized exception we wouldn't be testing that.

grzuy:
@mtrudel I think this should fix it https://github.com/mtrudel/bandit/pull/459 :pray:

PR #459, opened 2025-01-18
test: avoids test warnings about obvious but intended errors

grzuy:


mtrudel:
LGTM!

test/bandit/http1/plug_test.exs
```diff
@@ -58,7 +58,7 @@ defmodule HTTP1PlugTest do
     end
 
     def unknown_crasher(_conn) do
-      1 / 0
+      1 + self()
```
grzuy, line 61: Heads up, realized that Elixir 1.18 with it's smarter compiler and "type checker" will warn about this also...

mtrudel: I've fixed this by calling `apply(:erlang, :+, [1, self()])`

I also updated the other crash site to use this same call, previously it was using `1 = self()`. I assume that change is fine? They're both pushed to main in anticipation of a release later today

grzuy: > I've fixed this by calling `apply(:erlang, :+, [1, self()])`
> 

Hehe, cool.

> I also updated the other crash site to use this same call, previously it was using `1 = self()`. I assume that change is fine? 

Yeah, that's fine.
Thank you!

PR #475, opened 2025-03-14
Emit already_sent event in adapter

josevalim:
Future Plug versions will no longer emit the event automatically.


josevalim:
Thank you for double checking me. That was clearly a mistake. :D

Issue #492, opened 2025-05-16
Bandit ceases response mid stream


ns-blee:
We are proxying connections. When we set the initial window size (2MB) we are running into the following problem. It appears Bandit is not honoring the window size. It sends ~65.54k (the default) of data truncating at the exact same point then the connection is closed after 15 seconds. 

```
[F][ 0.001] send SETTINGS frame <length=18, flags=0x00, streamid=0>
     (niv=3)
     [SETTINGS_MAX_CONCURRENT_STREAMS(0x03):100]
     [SETTINGS_HEADER_TABLE_SIZE(0x01):4096]
     [SETTINGS_INITIAL_WINDOW_SIZE(0x04):2097152]
```

```
1	0.000000	163.116.194.127	10.122.140.186	TCP	68	34264 → 4010 [SYN] Seq=0 Win=65535 Len=0 MSS=1460 SACK_PERM=1 WS=64
2	0.000040	10.122.140.186	163.116.194.127	TCP	68	4010 → 34264 [SYN, ACK] Seq=0 Ack=1 Win=65535 Len=0 MSS=8960 SACK_PERM=1 WS=128
3	0.000236	163.116.194.127	10.122.140.186	TCP	56	34264 → 4010 [ACK] Seq=1 Ack=1 Win=65536 Len=0
4	0.000329	163.116.194.127	10.122.140.186	TLSv1.3	573	Client Hello
5	0.000341	10.122.140.186	163.116.194.127	TCP	56	4010 → 34264 [ACK] Seq=1 Ack=518 Win=66688 Len=0
6	0.004678	10.122.140.186	163.116.194.127	TLSv1.3	4888	Server Hello, Change Cipher Spec, Application Data, Application Data, Application Data, Application Data
7	0.004850	163.116.194.127	10.122.140.186	TCP	56	34264 → 4010 [ACK] Seq=518 Ack=4381 Win=74304 Len=0
8	0.004864	163.116.194.127	10.122.140.186	TCP	56	34264 → 4010 [ACK] Seq=518 Ack=4833 Win=77248 Len=0
9	0.005283	163.116.194.127	10.122.140.186	TLSv1.3	136	Change Cipher Spec, Application Data
10	0.005288	10.122.140.186	163.116.194.127	TCP	56	4010 → 34264 [ACK] Seq=4833 Ack=598 Win=66688 Len=0
[...]

PR #493, opened 2025-05-22
http2: merge settings frames

ns-blee:
Solution to: https://github.com/mtrudel/bandit/issues/492


mtrudel:
Closing in favour of #494, which just makes the settings field in SETTINGS frames be a simple map instead of a struct. Less code and less logic.

PR #517, opened 2025-08-21
Update version to `1.8`  in README

JoeARO:
Very minor PR but when swapping out `Plug.Cowboy` for `Bandit` we were only pulling in version `1.6.11` when the including `{:bandit, "~> 1.0"}` in our `mix.exs`:
Screenshot 2025-08-21 at 09 49 42

However `{:bandit, "~> 1.8"}` ensured we had the correct version:
Screenshot 2025-08-21 at 09 50 54

I haven't updated it in the `lib/bandit/phoenix_adapter.ex` documentation as the latest version of Phoenix seems to use `1.6.7`:
https://github.com/phoenixframework/phoenix/blob/v1.8.0/mix.lock#L2

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ Using Bandit to host your Phoenix application couldn't be simpler:
 1. Add Bandit as a dependency in your Phoenix application's `mix.exs`:
 
     ```elixir
-    {:bandit, "~> 1.0"}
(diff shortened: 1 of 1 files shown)


mtrudel:
Either way, we can merge this, there's no harm to it.

Thanks for the PR!

PR #520, opened 2025-09-01
Max concurrent streams

### LINKED ISSUE #519: Support for `SETTINGS_MAX_CONCURRENT_STREAMS`
I've noticed that currently this flag is not respected and infinite streams can be opened independently of configuration. I started having a look at the code and tests but, before I go too far into proposing a PR, I wanted to ask, is there any special reason why is so or would a contribution be welcomed? :)

@mtrudel

NelsonVides:
Closes #519, couldn't wait to get it done :)

Also adds protection against a rapid reset vulnerability, see
- https://blog.cloudflare.com/technical-breakdown-http2-rapid-reset-ddos-attack/
- https://nvd.nist.gov/vuln/detail/CVE-2023-44487
- https://github.com/ninenines/cowboy/pull/1617

Also, I've just noticed that when a rst_stream frame is received, the handler just sends an erlang process to the stream process, but, for HTTP2, this process is probably just running the Plug pipeline, and it is therefore never interrupted. Shouldn't it be sending exit signals to make sure they're dropped?

### DIFF
--- a/lib/bandit.ex
+++ b/lib/bandit.ex
@@ -164,6 +164,10 @@ defmodule Bandit do
     continuation frames. Defaults to 50_000 bytes
   * `max_requests`: The maximum number of requests to serve in a single
     HTTP/2 connection before closing the connection. Defaults to 0 (no limit)
+  * `max_cancel_stream_rate`: The maximum rate of stream cancellations (RST_STREAM frames) allowed.
+    Specified as a tuple of `{count, milliseconds}` where `count` is the maximum number of
+    RST_STREAM frames allowed within the time window of `milliseconds`. Defaults to `{500, 10_000}`
+    (500 cancellations per 10 seconds). Setting this to `nil` disables rate limiting
   * `default_local_settings`: Options to override the default values for local HTTP/2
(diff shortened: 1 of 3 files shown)


mtrudel:
#522 just got merged to main!

mtrudel:
With #524 having gone out, I think the only remaining piece here is the rapid reset mitigation? I'm a little anxious about that part in particular, mostly because it puts timing considerations on the hot path, even moreso a hot path that's shared across all streams in the connection. 

If you're able to, could you focus those changes in this PR? Now that #522 and #524 are both in main, this should be pretty straightforward I hope. Once there's a focused bit of work here we can start reasoning about rapid reset (I think, perhaps, the best path forward may be to make it opt-in / opt-out via configuration, but let's wait to see what the implementation looks like before adding any more bells and whistles).

mtrudel:
With #533 landing, this can close, correct?

NelsonVides:
Yes, by now all pieces from this one were merged in other places (#522, #524, and #533 for the record) 🎉 

Thanks a lot for your effort, appreciated! 🙂

PR #521, opened 2025-09-02
Use raw mode to read from the file info

NelsonVides:
See that Plug.Static already does this:
https://github.com/elixir-plug/plug/blob/87ae72ea2e21dd580df1367bdc709a84f75b87bd/lib/plug/static.ex#L137


NelsonVides:
Funny one, see that cowboy wasn't doing it either, just contributed to it too: https://github.com/ninenines/cowboy/pull/1688.

NelsonVides:
@mtrudel there's this tiny PR here around (and another much bigger one too) pending review :)

mtrudel:
Lint error is unrelated and mixed on main

PR #522, opened 2025-09-06
Pass stream_id to StreamError exceptions

NelsonVides:
Splitting from https://github.com/mtrudel/bandit/pull/520, this only adds the stream_id to the StreamError exceptions.


NelsonVides:
@mtrudel first one split here, this should barely be disruptive. Also added tests for it, which weren't there before in the other PR. There's only one failure, but I've tried running that exact elixir-otp match and it passes locally 🤷🏽‍♂️

PR #531, opened 2025-10-06
Add doc about EEx templates

achenet:
Add doc on how to use EEx templates with Bandit in README

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,46 @@ options via the `Bandit.start_link/1` function:
 Bandit.start_link(plug: MyPlug)
 ```
 
+## Using EEx Templates with Bandit
+
+[EEx](https://hexdocs.pm/eex/1.18.4/EEx.html) is the standard Elixir templating engine.
+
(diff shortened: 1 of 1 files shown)


mtrudel:
Thanks for the PR! I get the documentation gap you're trying to fill here, but it's not really part of Bandit's mandate to document how to use EEx; if anything that's more of a Plug concern, or perhaps even the EEx library itself.

Like any project, space in Bandit's README file is pretty sacred and should be laser focused on using the library and integrating it with its direct and necessary neighbouring libraries; anything more than that runs the risk of just overloading users. For that reason I'm not going to be merging this, but I'd suggest that you bring this forward as PR to either the [Plug library](https://github.com/elixir-plug/plug/) or the [EEx documentation](https://github.com/elixir-lang/elixir/blob/v1.12.3/lib/eex/lib/eex.ex#L10).

Regardless, thanks again for the PR!

achenet:
Ok, thanks for the feedback ^^

PR #532, opened 2025-10-20
Fix failing test in elixir 1.19

NelsonVides:
I don't know for what reason, this started failing in Elixir 1.19. It seems like `Kernel.inspect/1` prints differently by default now, cutting the string short much earlier. `Inspect.Algebra` has implemented changes in its algorithm that seem to affect this.

@mtrudel fixing the CI errors after merging last night's PR.


NelsonVides:
The autobahn failures are related to AWS being down this morning, as dockerhub was down with it 👀

mtrudel:
Thanks for the PR!

PR #540, opened 2025-11-16
Configurable deflate options for websockets

### LINKED ISSUE #527: deflate compression level for websocket servers
Hello,

I'm having trouble configuring the compression level for Bandit.WebSocket.PerMessageDeflate.  Is there an example of this?

When I set it in http_options like the docs suggest, it doesn't seem to make its way all the way into the per-message deflate negotiation.

Setting it in websocket_options doesn't pass the option validation.

Sorry in advance if I've overlooked anything

proxima:
Happy to add tests but would need a little guidance, I wasn't able to locate similar tests to emulate.

I didn't see any explicit validation of the deflate_options -- I'm guessing it's ok for zlib to choke on bad inputs?

Verified the options make it into PerMessageDeflate.init when using WebSockAdapter.upgrade(..., `compress: true`)

Closes #527

### DIFF
--- a/lib/bandit.ex
+++ b/lib/bandit.ex
@@ -205,7 +205,7 @@ defmodule Bandit do
         ]
 
   @typedoc """
-  Options to configure the deflate library used for HTTP compression
+  Options to configure the deflate library used for HTTP and WebSocket compression
   """
   @type deflate_options :: [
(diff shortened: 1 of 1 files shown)


mtrudel:
Thanks for the PR! Apologies it took so long for me to get to it!

PR #555, opened 2025-12-26
Add `response_encodings` option to http_options

### LINKED ISSUE #554: Enabling zstd
Was there ever any resolution on the discussion from #514 regarding configuring zstd compression?

I'd like to trial it on my site, but with the current implementation it will almost never be selected.

Thanks!

mtrudel:
Follow on from #514 
Fixes #554

### DIFF
--- a/lib/bandit.ex
+++ b/lib/bandit.ex
@@ -99,6 +99,9 @@ defmodule Bandit do
   * `compress`: Whether or not to attempt compression of responses via content-encoding
     negotiation as described in
     [RFC9110§8.4](https://www.rfc-editor.org/rfc/rfc9110.html#section-8.4). Defaults to true
+  * `response_encodings`: A list of compression encodings, expressed in order of preference.
+    Defaults to `~w(deflate gzip x-gzip zstd)`, with `zstd` only being present on platforms which
+    have the zstd library compiled in
   * `deflate_options`: A keyword list of options to set on the deflate library. A complete list can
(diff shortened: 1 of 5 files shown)


Fudoshiki:
Maybe `:accept_encodings` would be a more appropriate option by header name https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept-Encoding?

Fudoshiki:
Another option is allow to accept list in `:compress` 

```
compress: true
compress: ["zstd", "gzip"]
```

Fudoshiki:
> I've got 'compress' on my list of config values to deprecate in 2.0, mostly because it too is pretty non-specific.
> 
> 
> 
> Config naming aside, does this PR do what you wanted it to? Are you able to validate it in a deployed environment? 

yes

mtrudel:
Released as 1.10.0

oliver-kriska:
Hello,
just fyi after this was merged we had to do some changes in configuration of our app. We have `int_test` env which is used by cypress for e2e tests and after upgrade cypress was not able to connect properly with phoenix server. Everything else works. This config for that env helped:
```elixir
config :our_app, OurAppWeb.Endpoint,
  # Bandit 1.10 defaults to deflate-first which breaks wait-on/axios 
  http: [port: 4002, http_options: [response_encodings: ~w(gzip deflate x-gzip zstd)a]],
```

Fudoshiki:
same, always deflate

Issue #557, opened 2025-12-30
Default response_encodings change in 1.10.0 breaks HTTP client compatibility


neilberkman:
Bandit 1.10.0 changed the default `response_encodings` order from gzip-first to deflate-first. This breaks some HTTP client libraries used in dev tooling.

## Problem

The HTTP `deflate` encoding has interoperability issues due to ambiguity in the original spec - some implementations send/expect raw DEFLATE, others expect zlib-wrapped DEFLATE. Modern browsers handle this correctly, so real users are unaffected. The issue is with some HTTP client libraries.

## Impact

We discovered this when upgrading from 1.9.0 to 1.10.0. Our Cypress test suite uses `wait-on` (which uses the `got` HTTP client) to wait for the server. After the upgrade, `wait-on` would hang indefinitely because `got` couldn't properly decode the deflate-compressed response.

This is arguably a bug in `got` rather than Bandit - we may file there as well.

## Workaround

We're using `response_encodings: ~w(gzip deflate x-gzip zstd)a` in dev/test environments where tooling compatibility matters, and leaving the Bandit default in production since browsers work fine.

## Suggestion

Document this change in the changelog so users know to expect potential tooling breakage and can apply the workaround if needed.

## References

- https://stackoverflow.com/questions/388595/why-use-deflate-instead-of-gzip-for-text-files-served-by-apache
- https://zlib.net/zlib_faq.html#faq39

neilberkman:
Closing this - we discovered the actual issue was unrelated to deflate compression. Our server wasn't starting due to a different bug, and we incorrectly attributed it to the deflate change. After fixing the real bug, wait-on/axios works fine with Bandit's deflate-first default. Sorry for the noise.

mtrudel:
Would you be able to put together a tiny node project that could demonstrate this? I'd love to see if we're not doing something subtly wrong with deflate

pap:
### Additional Reproduction Case: Claude Code + Tidewave MCP

**Environment:**
- Bandit 1.10.0
- Phoenix 1.8.3
- Elixir 1.19.4 / OTP 28
- macOS Darwin 25.2.0

**Client:**
Claude Code v2.0.76 (Anthropic's Node.js-based CLI tool) connecting to Tidewave MCP server via HTTP transport.

**Observed Behavior:**
Claude Code fails to establish connection to Tidewave MCP endpoint with the following error:

  ZlibError fetching "http://localhost:4000/tidewave/mcp"

**Server Response:**
[...]

PR #562, opened 2026-01-06
Change precedence of compression orders

mtrudel:
Demote deflate, promote zstd in compression choices.

Relates to #557


mtrudel:
In 1.10.1

PR #565, opened 2026-01-24
feat: http2 sendfile streaming

elibosley:
Hi, 

I noticed that downloads do not stream, causing the entire file to be read from disk before the download can begin. This is fine for small files, but I have many very large files I'm managing that require much faster immediate resolution and send behavior. 

This adds configurable chunking to the sendfile endpoint to resolve the slow download setup, immediately sending headers for the download and starting it. In my testing, downloads started around 5-10x faster with an 8GB file from my remote server. 

Please let me know if I've missed anything or any required changes, I'm relatively new to elixir so I may not have the patterns down yet.


mtrudel (changes requested):
Minor change requested, but this is well done and a smart improvement!

mtrudel:
Published in 1.10.3!

PR #566, opened 2026-01-27
fix: detect client disconnect on timeout in ensure_completed

pepicrft:
When Bandit drains remaining body data after a controller returns (via `ensure_completed`), timeouts would always raise `HTTPError` even if the client had disconnected. This caused spurious error reports in error monitoring tools.

This is a follow-up to #564, which added client disconnect detection during body reading but did not cover the `ensure_completed` case.

## Changes

1. Extracts the client-disconnect detection logic into a reusable `handle_timeout_with_disconnect_check!/1` function
2. Applies the same detection in `ensure_completed`, converting timeouts to `TransportError` when the client has disconnected

This ensures that client disconnects during body draining are properly classified as `TransportError` (which can be filtered in error tracking) rather than `HTTPError`.

### DIFF
--- a/lib/bandit/http1/socket.ex
+++ b/lib/bandit/http1/socket.ex
@@ -316,27 +316,32 @@ defmodule Bandit.HTTP1.Socket do
           end
 
         {:error, :timeout} ->
-          # After a timeout, check if the peer is still connected. If not, this is
-          # likely a client disconnect that manifested as a timeout.
-          # We raise TransportError for disconnects and HTTPError for genuine timeouts.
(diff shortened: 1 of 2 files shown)


mtrudel:
Looks like a conflict - to avoid getting out of sync here, can you rebase this cleanly?

pepicrft (commented):
Done @mtrudel

mtrudel:
Thanks for the PR!

mtrudel:
Published in 1.10.3!

PR #576, opened 2026-03-17
Send 1000 on {:shutdown, :disconnected}

wwitek-whatnot:
## Background
We currently use Cowboy in our production system and tried switching to Bandit. This turned out to not be possible, as we started getting errors on client side, due to sockets being closed with code 1011.
After some investigation we found that this is because we broadcast "disconnect" to the id socket channel, in order to disconnect all user's sockets (as in [this](https://hexdocs.pm/phoenix/Phoenix.Socket.html#module-examples) example). In Bandit this goes to catch-all `{:stop, reason, websock_state}`, while in Cowboy, this goes through adapter [here](https://github.com/phoenixframework/websock_adapter/blob/1b0b2370e5e08b1474576bba34c991b6b5dbe977/lib/websock_adapter/cowboy_adapter.ex#L92) and so is then interpreted as a normal stop reason.
This is also very easy to reproduce by just generating two new phoenix apps with a single channel, one using cowboy and one using bandit. Then have a client join the channel, have an `:after_join` self-sent message trigger a disconnect, and observe the code on the client.

## Proposed change
We'd like to add `{:shutdown, :disconnected}` as one of the cases with 1000 code on close, as we believe this would fall under "socket being closed because it's purpose has been fulfilled" as defined in the [RFC](https://www.rfc-editor.org/rfc/rfc6455.html#section-7.4.1) and definitely doesn't fit 1011's "server is terminating the connection because it encountered an unexpected condition that prevented it from fulfilling the request".

### DIFF
--- a/.credo.exs
+++ b/.credo.exs
@@ -54,7 +54,7 @@
           #
           {Credo.Check.Refactor.Apply, []},
           {Credo.Check.Refactor.CondStatements, []},
-          {Credo.Check.Refactor.CyclomaticComplexity, [max_complexity: 15]},
+          {Credo.Check.Refactor.CyclomaticComplexity, [max_complexity: 16]},
           {Credo.Check.Refactor.FunctionArity, []},
           {Credo.Check.Refactor.LongQuoteBlocks, []},
           {Credo.Check.Refactor.MatchInCondition, []},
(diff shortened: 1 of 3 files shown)


mtrudel:
LGTM! There's a small credo error on code complexity coming out of this; I'll get it resolved and merged later today.

Thanks for the PR!

mtrudel:
Published as 1.10.4!

wwitek-whatnot:
> Published as 1.10.4!

Thank you!

PR #587, opened 2026-05-16
Elixir 1.20 warnings

tomciopp:
Fixes warnings emitted by elixir 1.20.0-rc.5.


mtrudel:
Thanks for the PR! Incredible the things that typing is catching!

PR #589, opened 2026-05-22
fix: keep websocket read timeout tied to reads

dl-alexandre:
## Summary
- track the timestamp of the last inbound WebSocket data
- keep Thousand Island read timeout scheduling relative to inbound reads, not local `handle_info/2` pushes
- add a regression test where periodic local pushes still end in a read timeout

## Tests
- `mix test test/bandit/websocket/sock_test.exs`

### DIFF
--- a/lib/bandit/websocket/handler.ex
+++ b/lib/bandit/websocket/handler.ex
@@ -24,6 +24,7 @@ defmodule Bandit.WebSocket.Handler do
       state
       |> Map.take([:handler_module])
       |> Map.put(:extractor, Extractor.new(Frame, primitive_ops_module, connection_opts))
+      |> Map.put(:last_read_at, monotonic_now())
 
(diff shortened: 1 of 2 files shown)


mtrudel:
What's the intent here? Is this a one-off fix to the underlying issue that's being addressed in https://github.com/mtrudel/thousand_island/pull/202?

dl-alexandre:
The intent was a narrow Bandit-side mitigation for the read-timeout behavior reported here, but looking at mtrudel/thousand_island#202, that seems like the better underlying fix because it separates the network timeout from GenServer message timeouts at the source.

If #202 is the intended direction, I'm happy to close this PR or re-scope it to only keep any Bandit-level regression coverage that would still be useful after Thousand Island is fixed.

mtrudel:
Thanks for the effort, but I think the Thousand Island fix in https://github.com/mtrudel/bandit/issues/202 is the path forward here.

PR #590, opened 2026-06-01
fix transfer-encoding

mize85:
added safe_downcase to headers -> "transfer-encoding" too
(Mac webdav sends "Chunked" instead of "chunked")


mtrudel (approved):
Change itself looks good. Can you add a quick test to cover the behaviour?

mize85:
@mtrudel  wow you are fast :)

sure, added one. 

one note:
can't reuse expect_tiny_chunked_body directly. that plug hard-asserts the raw header value is ["chunked"], and since Bandit only lowercases the header name (not the value), sending Transfer-encoding: Chunked makes the assertion fail with a 500.

Hope the dedicated expect_case_insensitive_chunked_body plug (which just reads the body) is ok to actually test the fix.

PR #596, opened 2026-06-04
fix(websocket): do not reset read timeout on outbound "sent" for WS connections

### LINKED ISSUE #577: WebSocket connection can survive silent half-open disconnect indefinitely
After migrating [asciinema-server](https://github.com/asciinema/asciinema-server/) from Cowboy to Bandit, I started seeing accumulation of stale websocket-connected LiveViews on production nodes.

The problem reproduces when a websocket connection becomes half-open in a way that does not deliver `FIN`/`RST` to the server, while the server continues to generate outbound websocket traffic for that connection.

In my case, periodic PubSub updates trigger LiveView diffs/pushes every 5 seconds. Those outbound/local messages appear to keep the Bandit websocket process alive indefinitely, so the transport never goes away and the attached LiveViews also remain alive.

This did not happen under Cowboy.

### Environment

- Elixir: 1.18
- Phoenix: 1.7.21
- Phoenix LiveView: 1.0.18
- Plug: 1.18.1
- Bandit: 1.10.2
- ThousandIsland: 1.4.3
- WebSockAdapter: 0.5.9

### Expected behavior

A websocket connection that has become silently half-open should eventually be reaped, even if the server is still generating outbound traffic for it.

At minimum, periodic server-side messages should not keep the connection alive forever in the absence of incoming client activity (e.g. LiveView heart beats).

### Actual behavior

If the server never receives `FIN`/`RST`, and the app continues to generate server-side websocket traffic for that connection, the websocket process remains alive indefinitely.

In asciinema server this leaves transport processes alive, and the LiveViews attached to those transports remain alive as well.

### Reproduction

I reproduced this on two machines on the same LAN:

- Linux workstation running asciinema server with Bandit
- MacBook running the browser client

The page under test mounts connected LiveViews that receive periodic updates every 5 seconds.

#### Steps

1. Start a Phoenix app using Bandit.
2. Open a page in the browser that mounts connected LiveViews and causes periodic server pushes over the websocket.
3. Confirm on the server that the websocket and LiveView processes are present (I used Phoenix LiveDashboard).
4. On the **server**, silently blackhole packets to/from the client IP for the websocket port, for example:

```bash
sudo iptables -I INPUT  -s <client-ip> -p tcp --dport 4000 -j DROP
sudo iptables -I OUTPUT -d <client-ip> -p tcp --sport 4000 -j DROP
```

This is intended to simulate a silent half-open connection where the server never sees FIN/RST.

5. After the drop is active, close the browser tab on the client.
6. Observe the server-side websocket / LiveView processes.

#### Result

After the packet blackhole is in place and the browser tab is closed on the client side, the websocket transport process on the server remains alive instead of being reaped. The LiveViews attached to that transport also remain alive. In this setup, periodic server-side updates continue to be generated, and the connection does not time out or terminate on its own.

In production, this shows up as a gradually increasing number of stale websocket-connected LiveViews on each node. As those accumulate, CPU usage rises together with memory usage until the node eventually hits its limits and the container is restarted, often due to OOM.

#### Control case

If I close the page normally without blackholing packets, cleanup happens quickly as expected.

dl-alexandre:
## Summary
A WebSocket connection that has become silently half-open (server can "send" but receives no FIN/RST or client frames) would stay alive indefinitely in Bandit if the application continued to generate outbound WS traffic for it (e.g. LiveView diffs every 5s via PubSub, or other periodic server pushes).

This did not happen under Cowboy.

## Root cause
In `Bandit.WebSocket.Handler`, `handle_info({:plug_conn, :sent}, ...)` did:
```elixir
{:noreply, {socket, state}, socket.read_timeout}
```
This rescheduled the ThousandIsland transport `read_timeout` on every outbound send.

Since the read_timeout is the mechanism for reaping idle connections (leading to `handle_timeout` -> close), continuous server sends kept pushing the timer out, even with zero client activity.

The WS layer has no separate "last client inbound activity" timer or mandatory ping/pong enforcement to probe the client independently of server traffic.

## Fix
Do not return/reset the `read_timeout` on the `{:plug_conn, :sent}` message for WS handlers. The timeout scheduled from the last actual client read (or initial) can now fire if the client goes silent, allowing the connection (and attached LV transports) to be reaped.

The `sent` reset appears to have been for HTTP/1 keepalive connection reuse; for long-lived WS it defeats client-driven idle detection.

## Validation
- Targeted change only in the WS sent handler.
- Existing ping/pong, close, error, and timeout paths in Connection/Handler unchanged.
- Gate will run compile/test/credo/etc.

Fixes mtrudel/bandit#577

### DIFF
--- a/lib/bandit/websocket/handler.ex
+++ b/lib/bandit/websocket/handler.ex
@@ -83,7 +83,13 @@ defmodule Bandit.WebSocket.Handler do
   def handle_timeout(socket, state), do: Connection.handle_timeout(socket, state.connection)
 
   def handle_info({:plug_conn, :sent}, {socket, state}),
-    do: {:noreply, {socket, state}, socket.read_timeout}
+    # Do not reset the read timeout on outbound "sent" for WS. This ensures that a
+    # silent half-open connection (no inbound client frames/heartbeats) will still
+    # be reaped by the transport read_timeout even if the server continues to
+    # generate outbound traffic (e.g. LV diffs, PubSub). Resetting on send was
+    # keeping such connections alive indefinitely (unlike Cowboy behavior).
+    # See mtrudel/bandit#577.
+    do: {:noreply, {socket, state}}
 
   def handle_info(msg, {socket, state}) do
     case Connection.handle_info(msg, socket, state.connection) do


mtrudel:
Thanks for the PR! 

There's actually a few other places where this is required; I've pushed them up in https://github.com/mtrudel/bandit/pull/597

PR #602, opened 2026-06-17
Add Bandit.HTTP2.Stream.send_trailers/2 for HTTP/2 trailer HEADERS support

gilbertwong96:
Adds a public `Bandit.HTTP2.Stream.send_trailers/2` function to send HTTP/2 trailer HEADERS frames after the response body ([RFC9113§8.1](https://www.rfc-editor.org/rfc/rfc9113.html#section-8.1)).

Trailers are commonly used by gRPC to carry `grpc-status` and `grpc-message` metadata at the end of responses ([gRPC over HTTP/2](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md)).

**What it does:**
- Validates trailers contain no pseudo-header fields ([RFC9113§8.2.2](https://www.rfc-editor.org/rfc/rfc9113.html#section-8.2.2)) and raises `Bandit.HTTP2.Errors.StreamError` on misuse
- Sends a HEADERS frame with `end_stream: true` via the existing `send/2` pattern
- Transitions the stream state to `:local_closed` (or `:closed` if the peer already closed)
- Exposed via `defdelegate` in the `Bandit.HTTPTransport` protocol implementation

**Tests (3 new):**
1. High-level: trailer round-trips through Req (status 200, body "body")
2. Low-level: server sends HEADERS + DATA + trailer HEADERS, verified at the wire via `SimpleH2Client`
3. Unit: pseudo-header fields in trailers raise `Bandit.HTTP2.Errors.StreamError`

**Quality:** ✅ 705 tests pass, 0 dialyzer errors, 0 credo issues, formatted clean.

### DIFF
--- a/lib/bandit/http2/stream.ex
+++ b/lib/bandit/http2/stream.ex
@@ -79,6 +79,47 @@ defmodule Bandit.HTTP2.Stream do
     }
   end
 
+  @doc """
+  Send an HTTP/2 trailer HEADERS frame on the given stream.
+  """
+  def send_trailers(%__MODULE__{state: state} = stream, trailers)
+      when state in [:open, :remote_closed] and is_list(trailers) do
+    validate_trailer_headers!(trailers, stream)
+
+    # Trailers always close the stream. The end_stream flag
+    # on a HEADERS frame terminates the stream (per
+    # RFC9113§6.2). The state transition mirrors the one in
+    # the defimpl's `set_state_on_send_end_stream/2` but is
+    # inlined here so we don't need to expose a new public
+    # function just for one caller.
+    send(stream.connection_pid, {{:send_headers, trailers, true}, stream.stream_id})
+    %{stream | state: state_after_send_end_stream(state)}
+  end
+
+  # State transition for the receiver of an end-stream signal
+  # (trailers always carry END_STREAM).
+  defp state_after_send_end_stream(:open), do: :local_closed
+  defp state_after_send_end_stream(:remote_closed), do: :closed
+
+  defp validate_trailer_headers!(trailers, stream) do
+    Enum.each(trailers, fn
+      {":" <> _, _} ->
+        raise Bandit.HTTP2.Errors.StreamError,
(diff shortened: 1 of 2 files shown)


mtrudel:
This looks good! I think I'll have some minor feedback but first I'd like to get this [surfaced up to Plug](https://github.com/elixir-plug/plug/issues/1121). 

Did you look at doing an accompanying HTTP/1 implementation?

gilbertwong96:
> This looks good! I think I'll have some minor feedback but first I'd like to get this [surfaced up to Plug](https://github.com/elixir-plug/plug/issues/1121).
> 
> Did you look at doing an accompanying HTTP/1 implementation?

I implement this feature only for HTTP/2 Stream because I'm trying to implement a gRPC server based on bandit and gRPC itself cannot be based on HTTP/1 protocol due to its streaming functions like bidi stream and streaming call. I don't think it's a good idea to implement `send_trailers/2` for HTTP/1 protocol.

It's not good to implement on the `plug` either because the `Plug.Conn` is a small finite-state machine with `state ∈ {:unset, :set, :set_chunked, :set_file, :file, :chunked, :sent, :upgraded}`. The state transitions assume "the response is sent" is a terminal event — once :sent, the connection is over. Trailers are a mid-response continuation: the response body has ended, but metadata is still being delivered. Plug has no slot for this.

mtrudel:
I'm afraid that I won't consider this for inclusion unless it comes with accompanying HTTP/1 support. I get that the primary use case is gRPC an thus only really relevant over HTTP/2, but the fact is that Bandit is a general purpose HTTP server first and foremost, and generalized support for features across all HTTP versions is table stakes for this. 

I also really do think that this should be included in Plug; I've already greased the skids for this in https://github.com/elixir-plug/plug/issues/1121 and the extra state changes there would be fine to add. 

Don't get me wrong; I've been a longstanding proponent of trailer support in the stack (as witnessed by the comment history on that Plug issue), but we need to do this properly.

gilbertwong96:
Thanks for your explanation

PR #603, opened 2026-06-20
Cache HTTP/1 connection data

preciz:
Assisted-by: Codex CLI : GPT 5.5

Avoid repeated peername lookups on HTTP/1 keep-alive requests by caching immutable conn_data in handler state.

Turns conn_data computation from “once per request” into “once per HTTP/1 connection.”

Benchmarked with 250k keep-alive requests / 10 headers: ~5% faster on average.

### DIFF
--- a/lib/bandit/http1/handler.ex
+++ b/lib/bandit/http1/handler.ex
@@ -8,13 +8,22 @@ defmodule Bandit.HTTP1.Handler do
   def handle_data(data, socket, state) do
     transport = %Bandit.HTTP1.Socket{socket: socket, buffer: data, opts: state.opts}
     connection_span = ThousandIsland.Socket.telemetry_span(socket)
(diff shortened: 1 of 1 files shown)


Fudoshiki:
@mtrudel Hi, I think this is not a bad improvement

mtrudel:
Hi! Sorry for the long delayed response. 

I'm always up for perf improvements, and the general idea here (caching of unchanging connection info) makes sense to me. From an implementation perspective, it might be better to have this done in the (currently unimplemented for this handler) ThousandIsland.handle_connection callback, which could write the conn_data directly into the handler's state. The call sites in handle_data could just blindly refer to it within state, instead of having to pass through the pattern matching helpers.

preciz:
@mtrudel like this?

mtrudel:
yeah exactly!

mtrudel:
published in 1.12.3

PR #606, opened 2026-06-29
Coalesce HTTP/2 response HEADERS+DATA into a single socket write

NelsonVides:
HTTP/2 wrote the HEADERS frame and DATA frame as two separate `ThousandIsland.Socket.send/2` calls (two TLS records) per response. This buffers the headers and flushes them with the first DATA frame in one write — the same trick HTTP/1 already uses via `send_buffer`.

**Result:** 2.0 → 1.0 socket writes per response; **~+30% req/s on small-body HTTP/2 over TLS** (`h2load`, loopback: tiny GET ~149k → ~197k median; 1/10 KiB similar; echo +~15%). This idea comes from https://erlangforums.com/t/livery-high-performance-http-1-1-http-2-http-3-server-for-erlang-otp-27/5693/5.

Second commit: a stream the client RST_STREAMs no longer gets an error response written to it, so Bandit now sends **zero** frames on a reset stream per RFC 9113 §6.4 (matches HTTP/1).

### DIFF
--- a/lib/bandit/http2/connection.ex
+++ b/lib/bandit/http2/connection.ex
@@ -319,7 +319,7 @@ defmodule Bandit.HTTP2.Connection do
     |> Enum.reduce(connection, fn pending_send, connection ->
       connection = connection |> Map.update!(:pending_sends, &List.delete(&1, pending_send))
       {stream_id, rest, end_stream, on_unblock} = pending_send
-      send_data(stream_id, rest, end_stream, on_unblock, socket, connection)
+      send_data(stream_id, [], rest, end_stream, on_unblock, socket, connection)
     end)
   end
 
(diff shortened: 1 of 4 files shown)


NelsonVides:
@mtrudel some CI failures seem related to CI flakiness or so. Have a look at the PR, it gets a nice perf improvement :)

lib/bandit/http2/connection.ex
```diff
@@ -341,50 +341,85 @@ defmodule Bandit.HTTP2.Connection do
...
+    {prefix_iodata, connection} =
+      if headers == [],
+        do: {[], connection},
+        else: encode_headers(stream_id, headers, false, connection)
 
-      if byte_size(rest) == 0 do
-        on_unblock.()
+    {data_iodata, rest, connection} = split_window(stream_id, data, end_stream, connection)
+
+    if prefix_iodata != [] or data_iodata != [],
```
mtrudel: Can we use something like Bandit.SocketHelpers.iodata_empty? here?

lib/bandit/http2/connection.ex
```diff
@@ -341,50 +341,88 @@ defmodule Bandit.HTTP2.Connection do
...
+      if headers == [],
+        do: {[], connection},
+        else: encode_headers(stream_id, headers, false, connection)
+
+    {data_iodata, rest, connection} = split_window(stream_id, data, end_stream, connection)
+
+    if !Bandit.SocketHelpers.iodata_empty?(prefix_iodata) ||
+         !Bandit.SocketHelpers.iodata_empty?(data_iodata) do
+      _ = ThousandIsland.Socket.send(socket, [prefix_iodata, data_iodata])
+      :ok
```
mtrudel, line 372: Bogus

mtrudel:
Sorry for the delay getting this merged. 

Looks good! Minor change to use existing SocketHelper functions, and mixup of a dialyzer error (the underlying error is with ThousandIsland's typing; I'm gonna fix is separate from this PR)

Issue #222, opened 2023-09-04
Received unsupported RSV flags 1


nthouliss:
Hi, 

I made a switch to Bandit with Phoenix (version 1.7.8) and got the following error on some of my websocket connections:

`** (stop) {:deserializing, "Received unsupported RSV flags 1"}`

I'm currently deployed on fly.io and my endpoint config isn't anything unusual. 

```
  config :example_app, App.Endpoint,
    url: [host: host, port: 443, scheme: "https"],
    check_origin: [origin],
    http: [
      ip: {0, 0, 0, 0, 0, 0, 0, 0},
      port: port
    ],
    secret_key_base: secret_key_base
 ```

(I just changed the app name but that's the endpoint config).

The only thing that I could be doing differently than most set-ups is having my own serialiser on my websocket. I use message pack on my websocket messages so I need to use a customer serialiser. 

This is likely a configuration issue rather than a bug in the library but I'm reporting since I'm not really sure where to start in fixing it. 

Thanks, happy to provide any additional details.

ryanwinchester:
Just a few questions to help make things super clear:

> I made a switch to Bandit with Phoenix (version 1.7.8)
1. You upgraded Phoenix and switched to Bandit at the same time?
2. Which version of Phoenix were you using previously?
3. It was deployed to fly.io, and working (with Cowboy and earlier Phoenix)?
4. Can you reproduce the error when running locally (on your machine)?

nthouliss:
> You upgraded Phoenix and switched to Bandit at the same time?

Yes, I upgraded at the same time. I've since rolled back the Bandit adapter and am running Phoenix 1.7.8 in fly.io with Cowboy without this issue. I've also updated my Elixir (now 1.14.5) and OTP (now 25.3.2.5) version. 

> Which version of Phoenix were you using previously?

I was using version 1.6.10 with Elixir 1.13.4 and OTP 25.0.2

> It was deployed to fly.io, and working (with Cowboy and earlier Phoenix)?

Yes, it was deployed and working before the change. I also tested it on a stage environment in fly.io beforehand (with Bandit) and didn't run into this issue. For some reason and only for some connections, this error occurs. 

> Can you reproduce the error when running locally (on your machine)?

I haven't been able to reproduce this issue locally.

moogle19:
This is odd.

The suspicion of @ryanwinchester seems correct. 
[...]

PR #610, opened 2026-07-09
Quiet websocket frame deserialization errors

ericmj:
When a client sends bytes that fail frame deserialization (invalid RSV flags, unknown opcodes, oversized frames), the websocket handler currently returns `{:error, {:deserializing, reason}, state}` and the connection process stops with that reason. Since it is not shutdown-shaped, OTP emits a "GenServer terminating" crash report for a condition entirely outside the server's control. In practice this is mostly scanners and fuzzers hitting public websocket endpoints with garbage; on hex.pm these crash reports were noisy enough that we filter them out of error tracking (hexpm/hexpm#1693).

This mirrors how HTTP protocol errors are already treated. The handler now calls `Connection.handle_error/3` inline, so the close frame (1002, or 1009 for oversized frames), websock `terminate/2`, and telemetry are unchanged. It then logs a one-line `** (exit) {:deserializing, ...}` summary governed by a new `log_protocol_errors` websocket option (defaulting to `:short`, with `false` disabling it, mirroring the option of the same name in `http_options`), and exits with `{:shutdown, {:deserializing, reason}}` so no crash report is emitted.

Related: #222 reported the same crash report shape for RSV flag errors, though that case looked like a real client negotiation problem rather than garbage traffic.

### DIFF
--- a/lib/bandit.ex
+++ b/lib/bandit.ex
@@ -208,6 +208,10 @@ defmodule Bandit do
   * `max_inflate_ratio`: The maximum allowable ratio to allow decompression of received WebSocket
     messages. Intended to prevent 'inflate bomb' attacks where a tiny deflated messages inflates to
     a massive one. Defaults to `25` representing a 25:1 allowable inflation ratio.
+  * `log_protocol_errors`: How to log protocol errors such as malformed frames. `:short` and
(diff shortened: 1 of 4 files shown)


mtrudel:
Apologies for the tardy review. Looks great! It'll go out with 1.12.2 later today

PR #613, opened 2026-07-14
Advertise HTTP/1 close at max_requests boundary

joshdchang:
## Summary

When an HTTP/1 connection reaches `max_requests`, Bandit serves the final allowed response and then closes the socket. That response currently does not advertise `Connection: close`, so a pooled client can reuse the connection before it observes the FIN and receive EOF instead of a response.

This change marks the final allowed request before running the Plug pipeline and forces exactly one `connection: close` header on its final non-informational response. The existing request lifetime remains bounded, `max_requests: 0` remains unlimited, and HTTP/2 behavior is unchanged.

## Verification

- Added a protocol regression test covering responses before and at the `max_requests` boundary
- `mix test`: 702 passed, 1 skipped, 3 excluded
- `mix credo --strict`
- `mix dialyzer`: 0 errors

### DIFF
--- a/lib/bandit/http1/handler.ex
+++ b/lib/bandit/http1/handler.ex
@@ -6,7 +6,13 @@ defmodule Bandit.HTTP1.Handler do
 
   @impl ThousandIsland.Handler
   def handle_data(data, socket, state) do
-    transport = %Bandit.HTTP1.Socket{socket: socket, buffer: data, opts: state.opts}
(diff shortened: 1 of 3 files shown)


mtrudel:
I'm trying to make a minor change to this PR but I'm unable to push to your branch. I'm gonna merge as-is and then make the change on main

PR #623, opened 2026-07-27
Pin GitHub Actions to commit SHAs

Totara-thib:
Hi Mat, small drive-by contribution: this pins every action and reusable workflow reference across the workflows to a commit sha, versions kept as comments.

A tag like `@v1` is a movable pointer: whoever controls the action, or anyone who compromises it, can re-point it and your next run executes their code with the job's token. The spot that motivated this is `erlef/setup-beam@v1` in `hex_publish.yml`, which runs with your Hex publish secrets. Same pattern as the tj-actions/changed-files incident (CVE-2025-30066).

Two notes:

- Your `elixir-ci-actions` reusable workflows are pinned to their current main commit too. I know it is your own repo so the risk model is different, if you would rather keep those on `@main` for convenience, say so and I drop that part.
- Dependabot keeps working exactly as today: it understands sha pins with a version comment and will keep opening its bump PRs.

All shas were resolved from the upstream repos and cross checked against their tags: setup-beam `54075bc` is v1.24.1, checkout `3d3c42e` is v7.0.1, cache `55cc834` is v6.1.0, upload-artifact `043fb46` is v7.0.1, elixir-ci-actions `18c287f` is main as of today. No logic changes, only the `uses:` lines.

Found with the Plumber CLI (https://github.com/getplumber/plumber), verified on main. I can also open a PR which adds it to CI so this does not quietly drift back, that one is a bonus, this PR stands on its own.

### DIFF
--- a/.github/workflows/autobahn.yml
+++ b/.github/workflows/autobahn.yml
@@ -42,16 +42,16 @@ jobs:
       AUTOBAHN_CASES: ${{ matrix.case }}
     steps:
     - name: Checkout code
-      uses: actions/checkout@v7
+      uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
     - name: Setup Elixir
-      uses: erlef/setup-beam@v1
+      uses: erlef/setup-beam@54075bcc5e249e4758d363f27d099f55d843f124 # v1.24.1
       with:
         elixir-version: ${{ matrix.elixir }}
         otp-version: ${{ matrix.otp }}
     - name: Disable compile warnings
       run: echo "::remove-matcher owner=elixir-mixCompileWarning::"
     - name: Retrieve mix dependencies cache
-      uses: actions/cache@v6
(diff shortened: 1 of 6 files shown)


mtrudel:
>Dependabot keeps working exactly as today: it understands sha pins with a version comment and will keep opening its bump PRs.

thanks for the clarity!

mtrudel:
This bailed out once merged to main, i had to revert it. Still interested in this in principle!

Totara-thib:
> This bailed out once merged to main, i had to revert it. Still interested in this in principle!

I found the cause: your Actions allowlist (nice by the way) has the literal pattern `erlef/setup-beam@v1`, so the sha refs stopped matching it and GitHub refused to start anything using setup-beam. Change that entry to `erlef/setup-beam@*` (`Settings > Actions > General`) and the pins work, the sha does the real pinning anyway. Dependabot bumps are unaffected.

Once changed, reverting the revert should go green. Or I can reopen a fresh PR, whatever is easier.

mtrudel:
Yep, fixed with a revert revert. Thanks for the PR!

PR #626, opened 2026-07-29
Avoid per-header length/1 and option lookups when reading HTTP/1 headers

### LINKED ISSUE #614: do_read_headers! re-fetches opts and calls length/1 on every header line
While reading through the HTTP/1 code I noticed `do_read_headers!` in `lib/bandit/http1/socket.ex` does a couple of things per header line that only need to happen once per request:
- `Keyword.get(socket.opts.http_1, :max_header_length, 10_000)` on every loop iteration
- `Keyword.get(..., :max_header_count, 50)` plus `length(headers)` for every header line — the `length/1` call makes the count check O(n²) over the request

Hoisting the two lookups out of the loop and carrying an integer count instead is behavior-identical (test suite passes unchanged, boundary behavior for "too many headers" is the same) and comes out to roughly 9-10% on a parse-only microbenchmark on my machine (13.6 → 12.4 µs/req at 10 headers, 57.2 → 51.5 at 45), which translates to about +3% e2e req/s for a browser-ish request and +5% for a 45-header one over a single keep-alive connection.

I'd like to take this one myself — the diff is small (+10/−7 in socket.ex) and I already have it working locally, so if this sounds reasonable I'll open a PR. Would be curious what the benchmark workflow says too. This is my first contribution here, so let me know if I'm missing anything process-wise.

jisung-02:
Closes #614.

`do_read_headers!` currently does two redundant things on every header line:
- calls `length(headers)` to check against `max_header_count`, which is O(n²) over the number of headers
- re-fetches `max_header_length` / `max_header_count` from opts via `Keyword.get` on each iteration (2n + 4 lookups per request)

This PR keeps an integer count instead of calling `length/1`, and hoists the two option lookups out of the read loop by splitting the function into an entry clause and a loop clause. 
Boundary behaviour is unchanged: up to `max_header_count` headers are accepted and the next one raises `request_header_fields_too_large`. Test suite passes unmodified.

Measured on Linux (OTP 27 / Elixir 1.18.4, matching the benchmark workflow): the two changes are independent — the counter removes the quadratic term, the hoisting shrinks the linear term — and together give roughly 5-7% on parse time for a 45-header request (4-6% for a typical 10-header request), consistent across CPU-restricted conditions. More
conservative than the numbers I posted in the issue, which came from a noisier setup.

For transparency: the fix itself is mine; the benchmark harness used to verify it (call counts, reduction fitting, wall-clock sweeps) was written with AI assistance. Happy to share the scripts if useful.

### DIFF
--- a/lib/bandit/http1/socket.ex
+++ b/lib/bandit/http1/socket.ex
@@ -146,22 +146,27 @@ defmodule Bandit.HTTP1.Socket do
     defp resolve_request_target!(_request_target, _method),
       do: request_error!("Unsupported request target (RFC9112§3.2)")
 
-    defp do_read_headers!(%@for{} = socket, headers \\ []) do
+    defp do_read_headers!(%@for{} = socket) do
       packet_size = Keyword.get(socket.opts.http_1, :max_header_length, 10_000)
+      max_header_count = Keyword.get(socket.opts.http_1, :max_header_count, 50)
+      do_read_headers!(socket, packet_size, max_header_count, [], 0)
+    end
 
+    defp do_read_headers!(socket, packet_size, max_header_count, headers, header_count) do
       case :erlang.decode_packet(:httph_bin, socket.buffer, packet_size: packet_size) do
         {:more, _len} ->
           chunk = read_available!(socket.socket, socket.socket.read_timeout)
           socket = %{socket | buffer: socket.buffer <> chunk}
-          do_read_headers!(socket, headers)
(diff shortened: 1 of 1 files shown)


mtrudel:
Solid! Thanks for the PR!

mtrudel:
This will go out in the next release, likely 1.13.0 in a few weeks

jisung-02:
Thanks for reviewing! Glad it helped.

PR #627, opened 2026-07-30
Fix typos

jisung-02:
Found with codespell 2.4.2 (excluding `.git`, `mix.lock`, `assets`).
- `handle_continutation/3` → `handle_continuation/3` in `lib/bandit/websocket/connection.ex`
  - 8 occurrences: the definition, the argument name, and 6 call sites. It's a `defp` and every reference lives in the same file, so this is internal only, no API change.
- `erorrs` → `errors` in `CHANGELOG.md`.

Split into two commits so the rename can be reviewed on its own.

Left alone as false positives:
- `TE` / `te` — the HTTP/2 TE header name (codespell wants `the`/`be`).
- `test/support/sendfile_large` — lorem ipsum filler text.
- `pre-selected` in `lib/bandit.ex` — a spelling preference, not a typo.

`mix compile --warnings-as-errors`, `mix format --check-formatted`, and the WebSocket test suite all pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,7 @@
 * Improve internal HTTP/2 error handling to be better about closing stream/connection
 * Coalesce header and body in to single transport send (#606, thanks @NelsonVides!)
 * Improve compression behaviour when streaming an explicitly length delimited body (#605, thanks @elibosley!)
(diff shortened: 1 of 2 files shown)


mtrudel:
Nice!

PR #631, opened 2026-08-08
Enforce the RFC9112 chunk-size grammar and reject repeated transfer-encoding

euriconicacio:
RFC9112§7.1 defines `chunk-size` as `1*HEXDIG`. `Integer.parse/2` additionally accepts a leading `+` or `-`, and the grammar was only ever checked by whether that parse succeeded, so signed values are read as valid chunk sizes.

Three consequences, all reproducible against 1.12.4:

| chunk size | today | expected |
|---|---|---|
| `+10` | read as a 16 byte chunk | 400 |
| `-0` | parsed as `0`, taken for the terminating chunk | 400 |
| `-1` | `FunctionClauseError` in `read_exactly!/5` | 400 |

`-0` is the one worth a second look. It ends the body at a byte position no conforming parser would end it at, so anything after it is read as a new request on the same connection:

```
POST /echo HTTP/1.1
Host: localhost
Transfer-Encoding: chunked

-0

GET /somewhere-else HTTP/1.1
Host: localhost

```

That is one request written to the socket and two responses back. Note this is specific to 1.12.x: before the chunked reader was reworked in #585 and #621 a zero-length chunk did not have a dedicated terminating branch, so `-0` hit the generic path and raised. The rework is correct on its own terms; it just gave the pre-existing grammar leniency somewhere to land.

`-1` never reaches a size check at all — `read_exactly!/5` is guarded on a non-negative length, so it matches no clause and the connection dies with a stack trace instead of a 400.

The fix checks the grammar explicitly rather than inferring it from the parse.

---

Separately, and in the same code path: `transfer-encoding` is resolved with `get_header/2`, which returns the first match and ignores the rest. Message framing is decided by two headers and only one of them, `content-length`, got the exhaustive treatment in f2ca636. So:

```
POST /echo HTTP/1.1
Transfer-Encoding: chunked
Transfer-Encoding: identity
```

is served as though it said plain `chunked`. Per RFC9112§6.1 repeated headers combine into one comma-separated value, and RFC9112§6.3 then requires rejecting the request because chunked is no longer the final encoding. Since the only encoding accepted here is a bare `chunked`, any repetition is either unsupported or has chunked in a non-final position, so rejecting outright loses nothing. Reversing the two lines already returns 400 today, via the unsupported encoding path.

---

**On why this is a pull request and not a security advisory.**

`-0` produces a genuine request boundary disagreement, so I went looking for an exploit path before writing this up, and did not find one. I replayed all of the above through nginx 1.27, HAProxy 3.0, Caddy 2, Traefik 3.3 and Envoy 1.32, each proxying to the same origin over a pooled upstream connection, followed by benign requests on fresh client connections to check for an orphaned response left queued for a later client. All five reject every case at the edge; nothing reached the origin and nothing leaked. With Bandit directly exposed the same attacker sends both requests anyway, so there is no gain.

So: conformance defects and one unhandled crash, no demonstrated impact. Happy to move this to the security channel instead if you read the impact differently — I would rather ask than assume.

---

Each case has a test. Reverting `lib/` on this branch and running `test/bandit/http1/protocol_test.exs` gives 4 failures; with the change the file is 158/158.

### DIFF
--- a/lib/bandit/headers.ex
+++ b/lib/bandit/headers.ex
@@ -26,6 +26,25 @@ defmodule Bandit.Headers do
     end
   end
 
+  # Transfer-encoding is special-cased for the same reason content-length is:
+  # it determines message framing, so resolving it with get_header/2 (which
+  # returns the first match and ignores the rest) lets a request whose framing
+  # is ambiguous be read as though it were not. Per RFC9112§6.1 repeated
+  # transfer-encoding headers combine into one comma-separated value, and
+  # RFC9112§6.3 requires rejecting a request whose final encoding is not
+  # chunked. Since the only encoding accepted here is a bare 'chunked', any
+  # repetition is either unsupported or has chunked in a non-final position,
+  # and can be rejected outright.
+  @spec get_transfer_encoding(Plug.Conn.headers()) ::
+          {:ok, binary() | nil} | {:error, String.t()}
+  def get_transfer_encoding(headers) do
+    case Enum.filter(headers, &(elem(&1, 0) == "transfer-encoding")) do
+      [] -> {:ok, nil}
+      [{"transfer-encoding", value}] -> {:ok, value}
+      _ -> {:error, "multiple transfer-encoding headers (RFC9112§6.1, RFC9112§6.3)"}
+    end
+  end
+
   # Covers IPv6 addresses, like `[::1]:4000` as defined in RFC3986.
   @spec parse_hostlike_header!(host_header :: binary()) ::
           {Plug.Conn.host(), nil | Plug.Conn.port_number()}
--- a/lib/bandit/http1/socket.ex
+++ b/lib/bandit/http1/socket.ex
@@ -62,7 +62,7 @@ defmodule Bandit.HTTP1.Socket do
       {method, request_target, socket} = do_read_request_line!(socket)
       {headers, socket} = do_read_headers!(socket)
       content_length = get_content_length!(headers)
-      body_encoding = safe_downcase(Bandit.Headers.get_header(headers, "transfer-encoding"))
+      body_encoding = get_transfer_encoding!(headers)
       request_connection_header = safe_downcase(Bandit.Headers.get_header(headers, "connection"))
       socket = %{socket | request_connection_header: request_connection_header}
 
@@ -201,6 +201,13 @@ defmodule Bandit.HTTP1.Socket do
       case Bandit.Headers.get_content_length(headers) do
         {:ok, content_length} -> content_length
         {:error, reason} -> request_error!("Content length unknown error: #{inspect(reason)}")
+      end
+    end
+
+    defp get_transfer_encoding!(headers) do
+      case Bandit.Headers.get_transfer_encoding(headers) do
+        {:ok, transfer_encoding} -> safe_downcase(transfer_encoding)
+        {:error, reason} -> request_error!("Transfer encoding unknown error: #{inspect(reason)}")
       end
     end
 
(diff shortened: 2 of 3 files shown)


mtrudel:
Thanks for this! Looks great!

mpettibone:
@mtrudel 

Just something to consider since there might an opportunity to make the code simpler to maintain and more clear where the load bearing is happening..


## Summary

#631 fixed real bugs: signed chunk sizes slipped through `Integer.parse/2`, so `+10` read as a sixteen byte chunk, `-0` passed for the terminating chunk, and `-1` crashed in `read_exactly!/5` instead of returning a 400. It also closed the repeated transfer-encoding hole. None of that is in question here, and none of it changes.

What this patch changes is how the chunk-size grammar gets enforced. #631 validates the line twice, with a hand-rolled byte walk followed by the parse. The two checks overlap almost completely, and the overlap is the maintenance problem: a reader can't tell which check is actually load-bearing without working out the redundancy for themselves. This patch collapses the validator into one function in two clauses where every check earns its place, keeps the accepted language byte-for-byte identical, keeps all four of #631's regression tests untouched, and adds one new test for an edge that becomes implicit in the new shape.

Net effect on `lib/bandit/http1/socket.ex`: 25 lines out, 15 in, and about half of those 15 are comment. Three private functions become one.

## What changes

Before (as merged in #631):

```elixir
defp parse_chunk_size!(chunk_size) do
  with true <- hex_digits?(chunk_size),
       {chunk_size, ""} <- Integer.parse(chunk_size, 16) do
    chunk_size
  else
    _ -> request_error!("Unable to parse chunk size")
  end
end

defp hex_digits?(<<>>), do: false
defp hex_digits?(chunk_size), do: all_hex_digits?(chunk_size)

defp all_hex_digits?(<<>>), do: true

defp all_hex_digits?(<<digit, rest::binary>>)
     when digit in ?0..?9 or digit in ?a..?f or digit in ?A..?F,
     do: all_hex_digits?(rest)

defp all_hex_digits?(_chunk_size), do: false
```

After:

```elixir
defp parse_chunk_size!(<<digit, _::binary>> = chunk_size)
     when digit in ?0..?9 or digit in ?a..?f or digit in ?A..?F do
  case Integer.parse(chunk_size, 16) do
    {chunk_size, ""} -> chunk_size
    _ -> request_error!("Unable to parse chunk size")
  end
end

defp parse_chunk_size!(_chunk_size), do: request_error!("Unable to parse chunk size")
```

## Why the single guard is enough

The reasoning is short, and it now lives in the comment above the function rather than in nobody's head. RFC9112§7.1 defines chunk-size as `1*HEXDIG`. `Integer.parse/2` is close to that grammar but not equal to it: the only extra thing it will consume is a single leading `+` or `-`. A sign can therefore only ever appear in the first byte. The existing `{chunk_size, ""}` match already requires the parse to consume every byte of the string, which rules out trailing junk like `5g`. So the one remaining gap between what `Integer.parse/2` accepts and what the grammar allows is that first byte, and anchoring it to a hex digit with a guard closes the gap exactly. The empty string can't match `<<digit, _::binary>>`, so it falls through to the error clause along with everything else.

That last point matters because the accepted language really is exactly `1*HEXDIG`, not an approximation of it. The equivalence was checked mechanically, not just argued: both validators were run side by side over 41 handpicked hostile inputs (signed values, `0x` prefixes, embedded whitespace, unicode digits, stray high bytes, empty strings) and 200,000 randomized strings over an adversarial alphabet. Zero disagreements.

## Why it's cleaner and easier to maintain

**Every check is load-bearing.** In the new shape, the guard is what rejects signs, and the `{chunk_size, ""}` match is what rejects trailing garbage. Delete either one and a test fails. In #631's shape, once `hex_digits?` returns true, `Integer.parse/2` on a nonempty all-hex string cannot return anything but `{n, ""}`, so the second branch of the `with` is unreachable as a failure path. Dead failure paths are where refactoring accidents happen: someone eventually "cleans up" the check that was actually doing the work, or preserves the one that wasn't, and the type system won't catch it either way.

**The validation is in one place, at the same level as the parse.** The reader of `parse_chunk_size!` sees the whole grammar decision in one screen: shape of the input in the clause head, remainder check in the case. Before, the answer to "what does this accept?" was spread across three functions, and understanding the empty-string behavior required noticing that `hex_digits?(<<>>)` is `false` while `all_hex_digits?(<<>>)` is `true`. Those two clauses look contradictory until you realize one is the API and the other is the recursion base. That's exactly the kind of local puzzle that costs review time forever after.

**It's the idiomatic shape for this job.** "Accept inputs of this shape, reject everything else" is what multi-clause heads with guards are for in Elixir. The fallback clause reads as the grammar's rejection rule rather than as an `else` catch-all swallowing two different failure kinds.

**One pass instead of two.** The line is walked once by `Integer.parse/2` instead of once by the recursion and again by the parse. With `@max_chunk_size_byte_count` capping the line at 16 bytes this is worth nothing in performance, and that's not the argument. The argument is that a second pass implies a second opinion, and second opinions invite divergence. There is now exactly one definition of what a chunk size is.

**The comment explains the subtle part.** #631's comment documented the three bad inputs, which is history worth keeping and is kept. What it didn't document is why any given check is sufficient, which is the thing the next maintainer actually needs before touching this function. The rewritten comment carries the sufficiency argument (sign only in the first byte, full consumption required) alongside the history.

## What deliberately does not change

Behavior is identical, so this is a pure refactor from the wire's point of view. All four regression tests from #631 are untouched and still pass: the leading plus sign, the signed zero with the smuggling follow-up assertion, the negative size, and the doubled transfer-encoding header. The transfer-encoding handling itself isn't touched at all, including the slightly awkward `"Transfer encoding unknown error:"` message, because it mirrors the existing `"Content length unknown error:"` convention in `get_content_length!` and consistency wins over wording.

## What's added

One regression test. In #631, rejecting an empty chunk size was an explicit clause (`hex_digits?(<<>>), do: false`). In the new shape it's implicit in the pattern match. Behavior that moves from explicit code into a pattern's failure mode deserves to be pinned, so the test sends `;ext=value\r\n`, a chunk line whose size portion is empty once extensions are split off, and asserts the 400 with the usual "Unable to parse chunk size" error. This case was rejected before #631, is rejected by #631, and stays rejected here; the point is that it can now never regress silently.

## Verification

- Applies cleanly on current `main` (92d63e1) and directly on the #631 merge commit (cd22b90).
- `test/bandit/http1/protocol_test.exs`: 162 tests, 0 failures (includes #631's four tests plus the new one).
- Full suite: 774 tests with a single failure in the `:inet6` server binding test, which fails identically on an unmodified checkout in the same environment (no IPv6 available), so it's environmental and unrelated.
- `mix format --check-formatted` clean; credo reports nothing beyond two pre-existing TODO notes.
- Randomized equivalence check between the old and new validators: 200,041 inputs, zero divergence.

mtrudel:
@mpettibone Thanks for the note. I've got Claude to whip up a PR for this (#661), which will be merging shortly.

PR #632, opened 2026-08-08
Treat proxy-connection as a connection-specific header in HTTP/2

euriconicacio:
RFC9113§8.2.2 names `Proxy-Connection` alongside `Connection`, `Keep-Alive`, `Transfer-Encoding` and `Upgrade` as connection-specific, and requires that a message carrying one be treated as malformed. `no_connection_headers!/2` covers the other four but not `proxy-connection`, so a request carrying it is served rather than reset.

The change adds only the omission, plus a test alongside the existing hop-by-hop one.

One observation I did not act on: the same list also carries `proxy-authenticate`, `proxy-authorization` and `trailers`, which RFC9113§8.2.2 does not name — those come from the older RFC2616 hop-by-hop list. Dropping them would make Bandit accept requests it rejects today, which is a behaviour change rather than a fix, so I left them alone. Mentioning it in case the divergence is news rather than deliberate.

### DIFF
--- a/lib/bandit/http2/stream.ex
+++ b/lib/bandit/http2/stream.ex
@@ -203,7 +203,7 @@ defmodule Bandit.HTTP2.Stream do
     # specific cases by RFC9113§8.2.2. We check those cases in a separate filter
     defp no_connection_headers!(headers, stream) do
       connection_headers =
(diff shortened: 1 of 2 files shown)


mtrudel:
Nice catch!

PR #645, opened 2026-08-11
Keep track of body_length to avoid list iteration

moogle19:
This avoids iterating the improper list via `IO.iodata_length/1` in every `do_read_chunked_data!/4`


mtrudel:
Nice!

PR #646, opened 2026-08-14
Fix non-deterministic compile-time code

fhunleth:
The BEAM files for these two modules would change when compiled even
though their corresponding source files were the same. This was due to
maps having an unspecified traversal order. Sorting the map entries or
keys makes their compilation deterministic.

This helps Nerves minimize the number of changes that need to be sent
over the wire for delta firmware updates.


mtrudel:
Wild how many places that unspecified map traversal order crops up. Thanks for this!

PR #665, opened 2026-08-20
Honor Accept-Encoding q-values, the wildcard, and case-insensitive codings

mpettibone:
Hi @mtrudel thanks again for creating such great software.

Note: Claude helped me write up this change request, and assisted with the ~50 lines of code and ~100 lines of tests.

## What this fixes

Response compression negotiation currently matches the client's `Accept-Encoding` tokens with a plain string membership test over the comma-split header. Tokens are compared whole and case-sensitively, so any parameterized or non-lowercase token fails to match anything. The four broken cases below (and the two already-correct anchor rows) were all verified against unmodified main:

| Accept-Encoding sent by client | today | with this PR |
|---|---|---|
| `gzip, deflate` | gzip | gzip |
| `gzip;q=1.0` | no compression | gzip |
| `br;q=1.0, gzip;q=0.5` | no compression | gzip |
| `*` | no compression | first server preference |
| `GZIP` | no compression | gzip |
| `gzip;q=0, deflate` | deflate | deflate |

The fix parses `Accept-Encoding` per RFC 9110 section 12.5.3: q-values are honored (a coding with `q=0` is a refusal), the `*` wildcard admits codings the client did not list, and codings compare case-insensitively per section 8.4.1. Selection stays in the server's own preference order, which is the same policy as today and a valid one (the server chooses among the codings the client finds acceptable).

## Why this is a value add

Real clients send q-values. The most prominent example is Ruby's standard library: `Net::HTTP` sends `Accept-Encoding: gzip;q=1.0,deflate;q=0.6,identity;q=0.3` on every request by default, which means every stock Ruby client talking to a Bandit service silently receives uncompressed responses today. Any client using q-values to express preferences among brotli, zstd, and gzip hits the same behavior, and it is invisible when it happens: nothing errors, responses are simply larger than they need to be. Given how much care Bandit's compression stack already puts into doing the right thing (zstd support, preference ordering, vary handling, etag and no-transform awareness), the negotiation step deserves to speak full RFC 9110 as well; this PR is the missing 40 lines of that story.

## Standards and protocol basis

RFC 9110 section 12.5.3 (Accept-Encoding: q-values, the `*` wildcard, and the identity coding), section 12.4.2 (quality values), and section 8.4.1 (content codings compare case-insensitively). Malformed q-values are read leniently as acceptance, on the reasoning that a client which bothered to list a coding presumably accepts it; `identity` yields no compression exactly as before, since it never appears in the server's encoding list.

## Why this approach is pragmatic

The change is confined to `negotiate_content_encoding/2` in `Bandit.Compression` plus two small private helpers; no other module is touched and the negotiation call sites are unchanged. The parse builds on `Plug.Conn.Utils.list/1` exactly as the current code does, then splits each element on `;` and reads a q parameter if one exists. For the common browser header (`gzip, deflate, br, zstd`, no parameters) the added work over today's code is a single-element split and a downcase per token, a handful of small allocations on a path that already allocates the token list. Server preference order is preserved as the selection policy rather than sorting by client q-value, deliberately: it keeps behavior identical for every header that works today and keeps the diff small; a q-value-ordered policy would be a separate conversation.

One test-design point worth flagging for review: the default `response_encodings` list is platform-dependent (zstd joins it when OTP 28's `:zstd` module is present), so the wildcard tests pin an explicit preference order rather than asserting against the default. Without that, `negotiate("*") == "gzip"` passes on OTP 27 and fails on OTP 28, where the wildcard correctly admits zstd first; Bandit's CI matrix covers both.

## Performance

One parse of a short header per request carrying `Accept-Encoding`, in place of the existing membership scan over the same comma-split list. No change when the header is absent or when `compress: false`. Response paths are untouched.

## Size

Library code: 46 added lines including comments in `lib/bandit/compression.ex` (the parser is about 30 lines of code). Tests: 117 lines across two files.

## Regression testing

Fifteen tests. Unit coverage in `test/bandit/compression_test.exs` (11 tests): plain lists select in server preference order exactly as before; case-insensitive matching; q-values on listed codings; `q=0` refusals including `0.0` and `0.000` forms; the wildcard, wildcard refusal, and wildcard with explicit refusals; whitespace and case tolerance around the q parameter; lenient handling of malformed q-values; identity handling including `identity;q=0, gzip`; `compress: false`; and a configured `response_encodings` order.

End-to-end coverage in `test/bandit/http1/compression_negotiation_test.exs` (4 tests): a Ruby `Net::HTTP` style header produces a gzip response that gunzips back to the original body; an uppercase coding compresses; a full `q=0` refusal produces an uncompressed response end to end; and a plain browser-style list behaves byte-for-byte as before (the no-regression anchor).


mtrudel:
Nice! There's a dangling .diff file here that you'll need to remove

mtrudel:
Also, the tests for this should live in http1/protocol_test and http2/protocol_test (even though this is shared code it's preferred practice here to dupe the tests into each protocol). Take care to slot them in next to existing compression tests (ordered by where transfer-encoding appears in the RFCs)

mpettibone:
It should be updated now. Let me know if I did this correctly or not (sorry, I don't use GitHub as much since I'm the odd person who primarily uses fossil/sourcehut)

mtrudel:
Looks good to me. Thanks for the work!

PR #666, opened 2026-08-20
fix: detect stale connections after request body reads

### LINKED ISSUE #649: Violating using the returned conn after `read_body` makes bandit corrupt the next request
I found this when looking into https://github.com/jimsynz/davy/pull/1. So this first of all is a bug in davy, which violated this portion of the plug docs on `read_body/2`:

> Like all functions in this module, the conn returned by read_body must be passed to the next stage of your pipeline and should not be ignored.

However the result of that bug is quite nasty: It made bandits `Bandit.HTTPTransport.ensure_completed` drop content beyond the current request leaving the next request corrupt:

```
17:50:24.447 [error] ** (Bandit.HTTPError) Request line HTTP error: "ion: keep-alive\r\n"

17:51:02.045 [error] ** (Bandit.HTTPError) Request line HTTP error: ": 179\r\n"

17:52:18.131 [error] ** (Bandit.HTTPError) Request line HTTP error: ": 179\r\n"
```

Claude suggested that bandit might be able to guard against that: 

> Where I think Bandit *could* reasonably harden, though: it already performs the real socket reads as a side effect, 
so it's fully capable of keeping its own independent tally — "how many body bytes have I actually pulled off this socket 
for this request, across every `read_body` call, regardless of which resulting conn got kept or discarded." That ledger 
lives in the connection process either way. Cross-checking that tally against what the *final* conn claims before trusting 
`ensure_completed` to read N more bytes is something only Bandit can do (Davy has no visibility into it) — and it would 
turn this failure mode from "silently eat the front of the next request" into a loud, immediate, obvious error, which is a 
much better failure mode for a library to have. That's a legitimate defense-in-depth suggestion, distinct from "Bandit 
should have caught Davy's bug" — it wouldn't prevent the underlying mistake, just make it fail loudly instead of 
corrupting an unrelated future request.

Not sure if there's arguments against that, but wanted to leave this here for consideration.

```elixir
Mix.install([{:bandit, "~> 1.5"}])

defmodule StaleConnPlug do
  @behaviour Plug
  import Plug.Conn

  @impl true
  def init(opts), do: opts

  @impl true
  def call(conn, _opts) do
    _body =
      case read_body(conn) do
        {:ok, body, _conn} -> body
      end

    # Bug: responds with the original `conn`, not the one read_body/2
    # returned above.
    send_resp(conn, 200, "ok")
  end
end

port = 4332
{:ok, _} = Bandit.start_link(plug: StaleConnPlug, port: port)
Process.sleep(300)

body = """
name: value
name: value
name: value
name: value
"""

request =
  IO.iodata_to_binary([
    "POST / HTTP/1.1\r\n",
    "Host: localhost:#{port}\r\n",
    "Content-Type: text/plain\r\n",
    "Content-Length: #{byte_size(body)}\r\n",
    "Connection: keep-alive\r\n",
    "\r\n",
    body
  ])

{:ok, socket} = :gen_tcp.connect(~c"localhost", port, [:binary, active: false, packet: :raw])

# Send request #1 one byte at a time, forcing Bandit through its
# incremental "more data" buffering path instead of getting the whole
# request in a single read().
for <<byte <- request>> do
  :gen_tcp.send(socket, <<byte>>)
  Process.sleep(1)
end

IO.inspect(:gen_tcp.recv(socket, 0, 2000), label: "response #1")

# Send request #2 as a single, perfectly well-formed write on the same
# now-corrupted connection. Expected: a second 200 OK. Actual: a 400 Bad
# Request, with "Request line HTTP error" logged above.
:gen_tcp.send(socket, request)
IO.inspect(:gen_tcp.recv(socket, 0, 2000), label: "response #2")

Process.sleep(300)
```

mvanhorn:
Plug requires callers to continue with the `Plug.Conn` returned by `read_body/2`, but an application can mistakenly discard it and later respond with the original connection. Bandit currently accepts that stale adapter state even though the body read has already consumed bytes from the socket, so `ensure_completed/1` drains the body again and can consume the beginning of the next keep-alive request. Issue #649 provides a deterministic content-length reproduction that forces incremental socket reads and demonstrates the next request being parsed from corrupted data. Bandit should turn this misuse into an immediate, visible failure without relying on OS-specific socket byte counters.

## Summary

Extend `Bandit.Adapter`'s existing owner-process validation with a per-request binding token kept both in the adapter and in the owning process dictionary. Initialize the token when the adapter is built, and after every successful `read_req_body/2` call rotate it while returning the new token only in the updated adapter, so any discarded pre-read connection becomes observably stale. Have the existing adapter-operation validation reject a token that is no longer current before a stale connection can send a response or trigger body draining; this keeps the invariant shared by HTTP/1 and HTTP/2 while adding no transport-specific or OS-dependent accounting.

## Testing

A plug retains the connection returned from `read_body/2`, sends its response normally, and the keep-alive connection remains usable for the next well-formed request.
A plug reads an incrementally delivered content-length body but calls `send_resp/3` with the pre-read connection; Bandit raises and logs an explicit stale/unbound-connection error rather than silently completing the request with stale transport state.
After stale-connection misuse, the server closes the HTTP/1 connection, so a pipelined or subsequently sent request is never parsed from bytes incorrectly drained by `ensure_completed/1`.
Reusing an earlier connection for another body read is rejected by the same binding validation, while consecutive reads that thread each returned connection continue to work.

Fixes #649

### DIFF
--- a/lib/bandit/adapter.ex
+++ b/lib/bandit/adapter.ex
@@ -8,9 +8,11 @@ defmodule Bandit.Adapter do
 
   @behaviour Plug.Conn.Adapter
   @already_sent {:plug_conn, :sent}
+  @owner_ref_key {__MODULE__, :owner_ref}
 
   defstruct transport: nil,
             owner_pid: nil,
+            owner_ref: nil,
             method: nil,
             status: nil,
             content_encoding: nil,
@@ -24,6 +26,7 @@ defmodule Bandit.Adapter do
   @type t :: %__MODULE__{
           transport: Bandit.HTTPTransport.t(),
           owner_pid: pid() | nil,
+          owner_ref: reference() | nil,
           method: Plug.Conn.method() | nil,
           status: Plug.Conn.status() | nil,
           content_encoding: String.t(),
@@ -38,6 +41,9 @@ defmodule Bandit.Adapter do
         }
 
   def init(owner_pid, transport, method, headers, opts) do
+    owner_ref = make_ref()
+    Process.put(@owner_ref_key, owner_ref)
+
     content_encoding =
       Bandit.Compression.negotiate_content_encoding(
         Bandit.Headers.get_header(headers, "accept-encoding"),
@@ -47,6 +53,7 @@ defmodule Bandit.Adapter do
     %__MODULE__{
       transport: transport,
       owner_pid: owner_pid,
+      owner_ref: owner_ref,
       method: method,
       content_encoding: content_encoding,
       expect_continue: expect_continue?(headers, Bandit.HTTPTransport.version(transport)),
@@ -82,7 +89,8 @@ defmodule Bandit.Adapter do
           |> Map.update(:req_body_bytes, byte_size(body), &(&1 + byte_size(body)))
           |> Map.put(:req_body_end_time, Bandit.Telemetry.monotonic_time())
 
-        {:ok, body, %{adapter | transport: transport, metrics: metrics}}
+        adapter = rebind(%{adapter | transport: transport, metrics: metrics})
+        {:ok, body, adapter}
 
       {:more, body, transport} ->
         body = IO.iodata_to_binary(body)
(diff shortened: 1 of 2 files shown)


mtrudel:
We pushed PRs at the same time for this! https://github.com/mtrudel/bandit/pull/668 is mine - let's chat in the original issue #649 about pros/cons

Issue #630, opened 2026-08-05
Opt-in client-disconnect notification while a Plug is executing (HTTP/1)


houllette:
_Note: the following has been generated with the aide of an LLM, but I have read over everything and done what I can for due diligence before submitting._

## Summary

For HTTP/1, an application currently learns that its client has disconnected
only when it **writes** (`{:error, :closed}` from `chunk/2`, or the
`Bandit.HTTPError` raised since #359). During windows where the application
writes nothing — long-running work before the first byte, or pauses between
chunks — there is no way to learn the client is gone, and no way to build one
from outside Bandit (details below, including why the obvious workarounds are
unsound rather than merely awkward).

This is a feature request for an **opt-in** mechanism that delivers client
disconnection to the handler process as an ordinary message while the Plug is
executing, mirroring the shape HTTP/2 already has with
`{:bandit, {:rst_stream, error_code}}`.

Related prior art: #344 reported the observable symptom for SSE and was closed
via #359, which surfaces **send** failures — the write-path half. This request
is about the write-idle half, which #359 explicitly does not cover. In the
[forum thread](https://elixirforum.com/t/handling-client-disconnect-during-serversentevents-using-plug-with-bandit/57919/2) you noted that eager connection shutdown on HTTP/1 is impossible under a
single-process-per-connection model given Plug's single-function API, and that
HTTP/2's multi-process model shuts streams down eagerly. This proposal stays
inside that constraint: no eager shutdown, no process-model change — just an
optional message the busy process can consume when it chooses to.

## Use case

We run an LLM reverse proxy on Bandit: the Plug relays a slow upstream HTTP
response (often streaming, tens of seconds to minutes) back to the client.
Write-based detection works well mid-stream. It cannot help during three
common windows where nothing is written to the client:

1. non-streaming requests (nothing is written until the upstream completes),
2. long time-to-first-byte while the upstream queues,
3. long pauses between upstream chunks.

When a client aborts during such a window, the handler keeps relaying (and the
upstream keeps metering) until the next write attempt or an idle timeout —
in our case up to 600 s — while the aborting client's retry consumes a second
upstream connection. Detecting the abort would let us cancel the upstream
request immediately, exactly as we already do when `chunk/2` returns
`{:error, :closed}`.

## Current behavior (bandit 1.12.4, thousand_island 1.5.0)

- Both Thousand Island transports hardcode `active: false`
[...]

Issue #490, opened 2025-05-09
More info on (Bandit.HTTPError) Body read timeout


ruslandoga:
👋 

I'm using Bandit v1.6.11 and I started noticing more 

```
[error] ** (Bandit.HTTPError) Body read timeout
```

in my logs today, like a couple every hour. And I wonder if there is a way to get more information on the client (like `inet:peername(Socket)` -- this would make it easier for me to ignore these errors if I see they are coming form DCs and not ISP IPs) or the state of the request (i.e. headers) from it?

I've tried attaching a handler that logs metadata on `[:bandit, :request, :exception]` but it wasn't called, and `[:thousand_island, :connection, :recv_error]` was called but didn't include enough information to aid in debugging this error:

```
[
  event: [:thousand_island, :connection, :recv_error],
  measurements: %{error: :timeout, monotonic_time: -576460688064789625},
  metadata: %{
    handler: Bandit.DelegatingHandler,
    telemetry_span_context: #Reference<0.1336770107.3872653314.19810>
  }
]
[error] ** (Bandit.HTTPError) Body read timeout
```

My handler:

```elixir
def attach_bandit_exceptions_logger do
  :telemetry.attach_many(
    "log-bandit-request-exceptions",
    [
      [:bandit, :request, :exception],
      [:thousand_island, :connection, :recv_error]
    ],
    &__MODULE__.log_bandit_request_exception/4,
    nil
  )
end

@doc false
def log_bandit_request_exception(event, measurements, metadata, _config) do
  IO.inspect(event: event, measurements: measurements, metadata: metadata)
end
```

mtrudel:
Body read timeouts are a protocol level concern, not a transport level concern, so from Thousand Island's perspective it's just a read that times out and thus completely 'normal' and not deserving of a `recv_error` event.

Conversely, Bandit's `:exception` event is intended to capture application level exceptions within your plug, and is not emitted on protocol level errors.

(Perhaps somewhat oddly), these are actually logged by Bandit as regular telemetry stop events [here](https://github.com/mtrudel/bandit/blob/b31059df6f9863fe5dda62228495b4a42348700f/lib/bandit/pipeline.ex#L223-L223), so if you hook on those (and match on an error field in the metadata) you should be able to capture these cases.

I'll also note that timeouts like this are very much an expected fact of life in network facing code, even in a intra-DC situation. Unless you know otherwise, there's not likely a lot of value in trying to chase these down

ruslandoga:
[...]

PR #669, opened 2026-08-20
Opt-in client disconnect notification for HTTP/1 plugs

mpettibone:
Hi @mtrudel 

Note: this is about ~100 lines of code and ~200 lines of test for this feature. Hope I'm not bombarding you with too many things today. Assisted by Claude. Please fill free to ask any questions.

## What this adds

A new HTTP/1 option, `http_1_options: [notify_disconnect: true]`. When enabled, Bandit watches the socket while your Plug is running, and the standard OTP socket close messages (`{:tcp_closed, _}` or `{:ssl_closed, _}`) arrive in the process executing the Plug if the client goes away. Since the exact message shape depends on the transport, the PR also adds one small public helper, `Bandit.client_disconnect_message?/1`, which recognizes disconnect messages across plain TCP, TLS, and HTTP/2 stream resets, so application code can write one `receive` clause that works everywhere.

A Plug serving server-sent events, relaying a slow upstream call, or doing a long computation can now notice mid-request that nobody is listening and wind down gracefully:

```elixir
def call(conn, _opts) do
  conn = send_chunked(conn, 200)

  receive do
    msg -> if Bandit.client_disconnect_message?(msg), do: raise("client went away")
  after
    30_000 -> :ok
  end

  # continue with the response...
end
```

## Why this is a value add

This PR aims to complete the disconnect handling by covering the one remaining window: a Plug that is computing or waiting, and not currently touching the socket. That window has grown in importance, because the workloads that live in it (server-sent events, LLM and upstream reverse proxies, long polls) have become everyday Elixir traffic. The tracker shows the same scenario coming up organically over time: [#344](https://github.com/mtrudel/bandit/issues/344), an [Elixir Forum thread](https://elixirforum.com/t/handling-client-disconnect-during-serversentevents-using-plug-with-bandit/57919) where your helpfully walked a user through the write-side signal, [#490](https://github.com/mtrudel/bandit/issues/490), and most recently [#630](https://github.com/mtrudel/bandit/issues/630) (Aug 2026), where a proxy author described relaying long upstream responses for clients that had already gone. There is clearly appetite for this, and your own incremental fixes suggest he cares about the scenario too; this PR just offers the next step.

On HTTP/2, the per-stream process model the you built in the 1.4.0 refactor gets this exactly right: a stream process is linked to its connection and sees `{:bandit, {:rst_stream, code}}` messages when a client cancels, so h2 plugs are never left working for a departed client. HTTP/1's single-process-per-connection design is a deliberate and successful choice (it is a big part of why Bandit's HTTP/1 engine benchmarks up to 4x Cowboy's, per the project's own CI), and the quiet tradeoff it makes is that nothing is reading the socket while a Plug runs. This PR keeps that design and its performance fully intact, and simply gives HTTP/1 plugs the same visibility their HTTP/2 siblings already enjoy, which also fits your stated principle from [PR #602](https://github.com/mtrudel/bandit/pull/602) that features should behave uniformly across HTTP versions. For what it's worth, Cowboy obtains the same property from its two-process-per-request model, at the cost of the extra process and message copying that Bandit's architecture intentionally avoids; this PR gets the property without paying that cost.

You mentioned once back in 2024 that "the single-function nature of the Plug API makes it impossible for a single-process-per-connection model such as Bandit to close connections eagerly." This PR agrees completely and does not attempt eager closing. Notification is a different and compatible thing: the disconnect becomes visible as a message, and what to do about it stays entirely in the application's hands, inside the Plug contract.

## Standards and protocol basis

This is server behavior rather than a wire-protocol feature, but it touches protocol correctness in two places, both handled. First, HTTP/1.1 connection reuse (RFC 9112 sections 9.3 and 9.6): any bytes the client sends while the socket is being watched are the start of the next pipelined request, and they are recovered into the read buffer so keepalive processing sees them exactly as if they had been read passively. Second, if a close or error was observed but the Plug did not consume the message, the connection is marked close-after-response so a finished connection is never reused. WebSocket upgrades (RFC 6455) are unaffected, covered by a regression test.

## Why this approach is pragmatic

The change is small because Bandit's architecture made it small. The `Bandit.HTTPTransport` protocol from the 1.4.0 refactor provides a clean seam for a cross-version concern, the option plumbing and buffer handling already exist, and the whole feature slots into them. The specific choices:

Opt-in, off by default. This changes the message-receiving environment of Plug processes (a `receive` loop written before this option existed could conceivably consume a socket message), and Bandit's performance discipline deserves protecting; nothing should be added to the default path. When off, the cost is one `Keyword.get` per request.

Arm only when there is nothing left to read. Erlang sockets in active mode and the passive `recv` calls used elsewhere in the module do not mix, so the socket is switched to `active: :once` only when the request body has been completely consumed, which is exactly when Bandit has no more reason to read. That happens at one of two points: at Plug invocation for bodyless requests (the SSE case), or the moment `read_data` consumes the final body byte, for both content-length and chunked bodies (the proxied POST case). This is why the option covers the #630 use case, where the Plug reads the body first and then waits on an upstream.

`active: :once`, never `active: true`. At most one data message can ever be delivered, so ordering between mailbox data and later passive reads stays trivially provable. The tradeoff is that notification is best-effort by nature: if a pipelined request arrives first and consumes the one shot, a subsequent close is discovered at the next socket touch, exactly as today. The docs say so plainly.

Disarm and recover before the response is committed. On the normal path, the pipeline returns the socket to passive mode and drains the mailbox: data messages are re-injected into the read buffer, close and error messages set `close_after_response`. The upgrade and error paths intentionally do not disarm; neither ever performs another passive read on the connection (the WebSocket handler consumes queued socket messages in arrival order through the standard active-mode path, and the error path closes), which is verified by tests.

Raw OTP messages plus a recognizer, rather than translated messages. The Plug runs in the handler process itself, so there is no intermediary that could rewrite `{:tcp_closed, _}` into something prettier while in flight. Instead of asking applications to know about `tcp_closed` vs `ssl_closed` vs h2's `{:bandit, {:rst_stream, _}}`, the `Bandit.client_disconnect_message?/1` helper owns that knowledge, and as a bonus gives HTTP/2 plugs the same contract with zero h2 changes.

The cross-version surface stays honest: `set_disconnect_notifications/2` is added to the internal `Bandit.HTTPTransport` protocol with a documented no-op for HTTP/2, since h2 stream processes already have both behaviors this option provides.

### DIFF
--- a/lib/bandit.ex
+++ b/lib/bandit.ex
@@ -139,6 +139,17 @@ defmodule Bandit do
     Defaults to 50 headers
   * `max_requests`: The maximum number of requests to serve in a single
     HTTP/1.1 connection before closing the connection. Defaults to 0 (no limit)
+  * `notify_disconnect`: Whether to notify the process executing your Plug when the client
+    disconnects while the Plug is still running. Once the request body has been completely read
+    (or if there is none), Bandit will watch the socket and standard `{:tcp_closed, _}` /
+    `{:ssl_closed, _}` messages will arrive in the process running your Plug if the client goes
+    away; `Bandit.client_disconnect_message?/1` can be used to match them. This is useful for
+    long-running requests such as server-sent events or proxied upstream calls, where the work
+    should be abandoned as soon as the client is gone. Notification is best-effort (it is
+    delivered while Bandit is between reads on the socket, which is the case for the entire time
+    your Plug is running after the body has been read). Plugs which use `receive` while this
+    option is enabled must be careful to not consume & discard these messages (or messages
+    starting with the tag of the next pipelined request). Defaults to `false`
   * `clear_process_dict`: Whether to clear the process dictionary of all non-internal entries
     between subsequent keepalive requests. If set, all keys not starting with `$` are removed from
     the process dictionary between requests. Defaults to `true`
@@ -154,6 +165,7 @@ defmodule Bandit do
           | {:max_header_length, pos_integer()}
           | {:max_header_count, pos_integer()}
           | {:max_requests, pos_integer()}
+          | {:notify_disconnect, boolean()}
           | {:clear_process_dict, boolean()}
           | {:gc_every_n_keepalive_requests, pos_integer()}
           | {:log_unknown_messages, boolean()}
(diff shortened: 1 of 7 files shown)


mtrudel:
Thanks for the work here, but as outlined in https://github.com/mtrudel/bandit/issues/630 I'm not really willing to entertain extending the shape of Bandit's Plug-facing API beyond what Plug provides. My view of the right path here is as I outlined on that issue.

mpettibone:
Sorry, didn’t see that. My mistake. Thanks for the review nonetheless.

mtrudel:
All good!

PR #672, opened 2026-08-21
Fix recent releases changelog headings in docs

dinokovac-river:
Screenshot 2026-08-21 at 10 26 52

It seems ex_doc / hexdocs doesn't link changelog headings in the sidebar if they are h1 instead of h2.

This change fixes it for recent changelog entries.


mtrudel:
Any reason not to go back all the way to the start of the changelog?

dinokovac-river:
> Any reason not to go back all the way to the start of the changelog?

Oh, I didn't notice initially - thanks.

I cleaned up the old version now as well, headings should be consistent now.

mtrudel:
Nice! thanks for the quality of life fix!

PR #679, opened 2026-08-25
RFC Conformance: Enforce Host requirements for HTTP/1.1 targets

mpettibone:

## Problem

Bandit validates the `Host` field only when the request-target does not already contain an authority. As a result, HTTP/1.1 absolute-form and CONNECT authority-form requests can omit `Host`, provide it more than once, or provide a value with an invalid port and still reach the Plug.

For example, this request is currently accepted:

```http
GET http://example.com/ HTTP/1.1

```

## RFC requirement

[RFC 9112 section 3.2](https://www.rfc-editor.org/rfc/rfc9112.html#section-3.2) requires a server to return `400 Bad Request` for an HTTP/1.1 request that has no `Host` field, more than one `Host` field line, or an invalid `Host` value. This requirement applies independently of the request-target form. [RFC 9110 section 7.2](https://www.rfc-editor.org/rfc/rfc9110.html#section-7.2) defines `Host` as `uri-host [ ":" port ]`, importing the host grammar from [RFC 3986 section 3.2.2](https://www.rfc-editor.org/rfc/rfc3986.html#section-3.2.2).

[RFC 3986 section 3.2.3](https://www.rfc-editor.org/rfc/rfc3986.html#section-3.2.3) permits an empty port, and [RFC 9110 sections 4.2.1 and 4.2.2](https://www.rfc-editor.org/rfc/rfc9110.html#section-4.2.1) assign the scheme default when the port is empty or omitted.

When an absolute-form request-target is present, the request-target authority still takes precedence when constructing the effective request URI. Validating `Host` does not change that routing rule.

## Implementation

The HTTP/1.1 host-selection path now always obtains and parses exactly one `Host` field before selecting the effective host and port:

- Origin-form requests continue to derive the host and port from `Host`.
- Absolute-form and authority-form requests continue to derive the effective host and port from the request-target.
- Missing, repeated, or unparsable `Host` fields now produce `400 Bad Request` for every HTTP/1.1 request-target form.
- Host parsing now enforces the RFC 3986 `reg-name`, IPv6, and IPvFuture grammar while accepting a syntactically empty port as the scheme default.
- HTTP/1.0 still does not require `Host`; when a Host field is supplied, it uses the same URI-host parser.
- Because the parser is shared with HTTP/2 `:authority`, malformed ordinary HTTP/2 authority syntax is converted to a stream `PROTOCOL_ERROR` rather than an HTTP/1-style error response.

## Tests

The protocol tests now verify that Bandit rejects:

- An absolute-form HTTP/1.1 request without `Host`.
- An absolute-form request with duplicate `Host` fields.
- An absolute-form request with an invalid `Host` port.
- Absolute-form requests with delimiters, userinfo, malformed percent-encoding, or an invalid bracketed IP literal in `Host`.
- A CONNECT authority-form HTTP/1.1 request without `Host`.

A positive regression test verifies that `Host: example.com:` uses the default HTTP port.

Focused parser tests cover registered names, percent-encoding, IPv6, IPvFuture, empty ports, and malformed authority syntax.

An HTTP/2 regression test verifies that an invalid ordinary `:authority` resets only its stream and leaves the connection usable.

Existing tests continue to verify that the absolute-form request-target authority takes precedence over a valid but different `Host` value.

- Full non-slow suite: 822 tests pass (the suite's one IPv6 server binding test needs an IPv6-capable host)

### DIFF
--- a/lib/bandit/headers.ex
+++ b/lib/bandit/headers.ex
@@ -102,38 +102,112 @@ defmodule Bandit.Headers do
     end
   end
 
-  # Covers IPv6 addresses, like `[::1]:4000` as defined in RFC3986.
+  # RFC9110§7.2 defines Host as uri-host [ ":" port ], importing uri-host from
+  # RFC3986§3.2.2. Bracketed hosts therefore have to be IPv6 or IPvFuture literals, while
+  # unbracketed hosts have to match reg-name (IPv4 addresses are also valid reg-names).
   @spec parse_hostlike_header!(host_header :: binary()) ::
           {Plug.Conn.host(), nil | Plug.Conn.port_number()}
-  def parse_hostlike_header!("[" <> _ = host_header) do
-    host_header
-    |> :binary.split("]:")
-    |> case do
-      [host, port] ->
-        case parse_integer(port) do
-          {port, ""} when is_port_number(port) -> {host <> "]", port}
-          _ -> raise Bandit.HTTPError, "Header contains invalid port"
+  def parse_hostlike_header!("[" <> rest) do
+    case :binary.split(rest, "]") do
+      [literal, suffix] ->
+        if valid_ip_literal?(literal) do
+          {"[" <> literal <> "]", parse_port_suffix!(suffix)}
+        else
+          raise Bandit.HTTPError, "Header contains invalid host"
         end
 
-      [host] ->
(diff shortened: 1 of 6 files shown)


mtrudel:
Nice!

PR #680, opened 2026-08-25
RFC Conformance: Complete HTTP/2 field validation

mpettibone:

## Problem

Bandit's HTTP/2 request validation rejects uppercase regular field names and values containing CR, LF, or NUL. It does not reject several other field-name and field-value forms that RFC 9113 defines as malformed.

Examples that can currently pass validation include:

```text
bad name: value
bad:name: value
x-test:  leading
x-test: trailing<TAB>
```

Field-value validation is also applied only to regular fields, leaving pseudo-field values outside the shared validation step.

## RFC requirement

[RFC 9110 section 5.1](https://www.rfc-editor.org/rfc/rfc9110.html#section-5.1) defines a field name as a non-empty `token`, using the grammar in [section 5.6.2](https://www.rfc-editor.org/rfc/rfc9110.html#section-5.6.2). [RFC 9110 section 5.5](https://www.rfc-editor.org/rfc/rfc9110.html#section-5.5) limits field values to visible ASCII or `obs-text`, with SP and HTAB allowed internally; other control bytes and DEL are outside the grammar. [RFC 9113 section 8.2.1](https://www.rfc-editor.org/rfc/rfc9113.html#section-8.2.1) adds HTTP/2-specific requirements and requires an endpoint to treat a message as malformed when:

- A field name contains bytes `0x00` through `0x20`, uppercase ASCII, or bytes `0x7f` through `0xff`.
- A regular field name contains a colon.
- A field value contains NUL, CR, or LF.
- A field value begins or ends with SP or HTAB.

For a malformed request, [RFC 9113 section 8.1.1](https://www.rfc-editor.org/rfc/rfc9113.html#section-8.1.1) requires the server to reset the stream with `PROTOCOL_ERROR`.

## Implementation

The HTTP/2 stream validator now:

- Rejects empty names and every byte outside the RFC 9110 `token` grammar, including delimiters, controls, spaces, DEL, and non-ASCII bytes.
- Preserves the existing, more specific uppercase-field diagnostic.
- Rejects control bytes and DEL outside the RFC 9110 field-value grammar while retaining legal `obs-text` and internal SP or HTAB.
- Rejects field values beginning or ending with SP or HTAB.
- Applies field-value validation to pseudo-fields as well as regular fields.
- Treats a field value that HPAX decodes as `nil` as the empty value it represents. HPAX decodes an indexed static-table entry with no value, such as a bare `user-agent` at index 58, to a nil value (elixir-mint/hpax#27). On current `main` such a request draws a 500 from an `ArgumentError` inside validation, and h2spec `generic/5/1` requires it to be accepted; regular fields are normalized to the empty binary before reaching the Plug.
- Performs field-value checks before interpreting pseudo-field values such as `:authority`.

The existing CR, LF, and NUL checks remain in place.

## Scope notes

`Bandit.Headers.field_value_valid?/1` also backs HTTP/1 header validation, so HTTP/1 requests whose field values contain C0 control bytes other than HTAB, or DEL, are now rejected with 400 as well. RFC 9110 section 5.5 places those bytes outside the field-value grammar for every HTTP version; rejecting them is mandatory for HTTP/2 and permitted for HTTP/1. CR, LF, and NUL were already rejected on both protocols.

The recursive byte-level validation replaces the cached compiled-pattern scan. Measured on typical short header values the recursive walk is faster than the previous `:binary.match/2` scan, and within a few percent on multi-kilobyte values, so the header hot path does not regress.

## Tests

New protocol tests send HPACK-encoded requests containing:

- Empty regular field names and names containing spaces, tabs, delimiters, DEL, or `0xff`.
- Control bytes and DEL embedded in regular field values.
- Leading and trailing SP in field values.
- Leading and trailing HTAB in field values.
- Invalid edge whitespace and control bytes in pseudo-field values.
- An indexed static-table field with no value (`user-agent` at index 58), which must be accepted and observed by the Plug as an empty value.

Each case asserts that Bandit sends `RST_STREAM` with `PROTOCOL_ERROR` while keeping the HTTP/2 connection usable.

- Full non-slow suite: 821 tests pass (the suite's one IPv6 server binding test needs an IPv6-capable host)
- HTTP/2 protocol suite: 184 tests pass

### DIFF
--- a/lib/bandit/headers.ex
+++ b/lib/bandit/headers.ex
@@ -2,33 +2,21 @@ defmodule Bandit.Headers do
   @moduledoc false
   # Conveniences for dealing with headers.
 
-  @invalid_field_value_pattern_key {__MODULE__, :invalid_field_value_pattern}
-
   @spec is_port_number(integer()) :: Macro.t()
   defguardp is_port_number(port) when Bitwise.band(port, 0xFFFF) === port
 
-  # RFC9110§5.5 / RFC9113§8.2.1: field values containing CR, LF, or NUL characters
-  # are invalid and dangerous (a common request-smuggling / response-splitting /
-  # log-injection vector). Shared by both HTTP/1 and HTTP/2 so that neither
-  # transport can drift from the other's validation. The match pattern is
-  # compiled once (lazily, on first use) and cached in :persistent_term, since
-  # compiling it on every call is a measurable fraction of header parsing time.
+  # RFC9110§5.5 defines field values in terms of VCHAR (0x21-0x7e), obs-text
+  # (0x80-0xff), and internal SP / HTAB. All other control bytes and DEL are
+  # outside the field-value grammar. HTTP/2 applies the additional edge-whitespace
+  # prohibition from RFC9113§8.2.1 in its field-section validator.
   @spec field_value_valid?(binary()) :: boolean()
-  def field_value_valid?(value) do
-    :binary.match(value, invalid_field_value_pattern()) == :nomatch
-  end
+  def field_value_valid?(<<>>), do: true
 
-  defp invalid_field_value_pattern do
-    case :persistent_term.get(@invalid_field_value_pattern_key, :undefined) do
-      :undefined ->
-        pattern = :binary.compile_pattern(["\r", "\n", "\0"])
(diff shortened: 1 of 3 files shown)


mtrudel:
Nice

PR #682, opened 2026-08-25
RFC Conformance: Enforce HTTP/2 trailer field section requirements

mpettibone:
I had Claude check the code against various RFC conformance, and it flagged below. Claude helped with the desc.

## Summary

Validate request trailers as an HTTP/2 field section and require the trailer HEADERS frame to carry `END_STREAM`.

## What was not conformant

Bandit accepted a second request HEADERS frame without `END_STREAM`, logged and ignored its fields, and continued reading DATA. It also checked only for pseudo-headers in trailers, so the normal HTTP/2 rules for lowercase names, field values, connection-specific fields, and `TE` were not applied.

[RFC 9113 section 8.1](https://www.rfc-editor.org/rfc/rfc9113.html#section-8.1) says a trailer field section must terminate the stream. It specifically requires `END_STREAM` on the terminating HEADERS frame and says a trailer section without it is a malformed request. [RFC 9113 section 8.2.1](https://www.rfc-editor.org/rfc/rfc9113.html#section-8.2.1), [RFC 9110 section 5.1](https://www.rfc-editor.org/rfc/rfc9110.html#section-5.1), and [RFC 9113 section 8.2.2](https://www.rfc-editor.org/rfc/rfc9113.html#section-8.2.2) define the field-name, field-value, and connection-specific field rules that apply to HTTP/2 field sections.

## Implementation

- Reject trailer HEADERS unless the frame carries `END_STREAM`.
- Continue rejecting pseudo-headers in trailers.
- Apply Bandit's existing lowercase-name, field-value, connection-specific field, and `TE` validators to trailers before discarding them.
- Enforce the full regular field-name token grammar and the HTTP/2 prohibition on edge SP or HTAB in trailer values.
- Validate every member of every `TE` field line, while accepting repeated `trailers` members.
- Match RFC 9113's exact connection-specific field list without rejecting unrelated extension fields by name.
- Keep the existing behavior of accepting and ignoring a valid trailer section.
- Treat a trailer field value that HPAX decodes as `nil` (an indexed static-table entry with no value) as the empty value it represents, instead of letting it crash validation (elixir-mint/hpax#27).

This is intentionally a focused change. It does not add trailer exposure to Plug and does not alter valid request-body handling.

## Tests

The protocol tests cover:

- acceptance of a valid trailer section with `END_STREAM`, including an unrelated extension field named `trailers`;
- rejection of a trailer section without `END_STREAM` with stream `PROTOCOL_ERROR`;
- rejection of a connection-specific trailer field;
- rejection when any of multiple `TE` field lines contains a value other than `trailers`;
- rejection of invalid trailer field names; and
- rejection of CR and edge whitespace in trailer field values; and
- acceptance of a trailer field encoded as an indexed static-table entry with no value.

Run:

```console
mix test test/bandit/http2/protocol_test.exs
```

- Full non-slow suite: 821 tests pass (the suite's one IPv6 server binding test needs an IPv6-capable host)
- HTTP/2 protocol suite: 184 tests pass

### DIFF
--- a/lib/bandit/http2/stream.ex
+++ b/lib/bandit/http2/stream.ex
@@ -219,7 +219,7 @@ defmodule Bandit.HTTP2.Stream do
     # specific cases by RFC9113§8.2.2. We check those cases in a separate filter
     defp no_connection_headers!(headers, stream) do
       connection_headers =
-        ~w[connection keep-alive proxy-authenticate proxy-authorization proxy-connection trailers transfer-encoding upgrade]
+        ~w[connection keep-alive proxy-connection transfer-encoding upgrade]
 
       if Enum.any?(headers, fn {key, _value} -> key in connection_headers end),
         do: stream_error!("Received connection-specific header", stream)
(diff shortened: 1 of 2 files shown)


mtrudel:
More / better trailers support remains my favourite thing.

PR #685, opened 2026-08-25
RFC Conformance: Accept empty non-final continuation frames

mpettibone:

## Summary

Accept zero-length non-final continuation frames while preserving Bandit's fragmented-message size protection and avoiding state growth for empty fragments. Re-enable Autobahn case 6.1.2.

## RFC conformance

[RFC 6455 Section 5.4](https://www.rfc-editor.org/rfc/rfc6455.html#section-5.4) defines a fragmented message as an initial non-final data frame followed by zero or more non-final continuation frames and a final continuation frame. The same section says that fragments of any size may be created by a sender and that receivers must support fragmented messages.

Bandit currently closes the connection with status 1008 when it receives a zero-length non-final continuation frame. That frame is valid according to the RFC. The deliberate rejection also requires Autobahn case 6.1.2 to be excluded from CI.

## Implementation

When Bandit receives an empty non-final continuation during a fragmented message, it now keeps the existing fragmentation state and waits for the next frame.

The empty frame is not appended to the accumulated iodata. This is important because an arbitrary number of empty continuations therefore does not grow `fragment_frame` or `fragment_size`. Non-empty continuations still use the existing `max_fragmented_message_size` checks.

Autobahn case 6.1.2 is restored to the CI case list.

## Tests

The protocol test now sends:

1. A non-final binary frame containing `AB`.
2. Three empty non-final continuation frames.
3. A final continuation containing `CD`.

It verifies that the connection remains open and the application receives the single reconstructed message `ABCD`.

Run:

```console
mix test test/bandit/websocket/protocol_test.exs
mix test --only slow test/bandit/websocket/autobahn_test.exs
```

- Full non-slow suite: 814 tests pass (the suite's one IPv6 server binding test needs an IPv6-capable host)
- The Docker-backed Autobahn suite is not runnable in the verification environment; case 6.1.2 is restored in the CI workflow so the project's own Autobahn job exercises it


mtrudel:
This was intentional, landed in https://github.com/mtrudel/bandit/commit/1e8e55966da9129016b73d32f0e1df4630e3b463 as part of the mitigation of (I believe; GitHub's security fork process is terrible at leaving a paper trail, oddly enough) https://github.com/mtrudel/bandit/security/advisories/GHSA-pf94-94m9-536p

I *think* that concern has been mitigated here by the fact that you don't accumulate empty values into the holist, but I'd ask that you go over the issue in the advisory above and evaluate this change in that light (I'll do the same).

mpettibone:
Ah - thanks for that. So I went back through GHSA-pf94-94m9-536p and 21612c7 before touching this further. The exposure the 1008 rejection addressed was that an empty non-final continuation still appended a cell to the fragment iodata, so a peer could grow per-connection state indefinitely without ever advancing fragment_size, which is the quantity max_fragmented_message_size bounds. 

This PR takes a different line: the empty fragment is a pure no-op and is never appended, so fragment state does not change at all and the limit's accounting stays exact. Per-frame cost is then the same as any other small frame a peer can already stream (a ping, say), bounded by read rate rather than accumulated.

I've pushed two additions to make that explicit: a comment at the site tying the no-op to the advisory, and a regression test asserting that handling an empty non-final continuation leaves fragment_frame and fragment_size untouched. I also noticed the branch was missing the workflow change this PR describes; .github/workflows/autobahn.yml now restores case 6.1.2 so the Autobahn job exercises exactly this behavior.

mpettibone:
Hold one moment. Something is odd about my patch to the test coverage, investigating now.

mpettibone:
Ok, I fixed the issue with test coverage. Should all be good to review now (sorry about that).

